Issue #2, opened 2023-05-03
TypeError: generate() got an unexpected keyword argument 'new_text_callback'


untszlung:
/privateGPT/gpt4all_j.py", line 152, in _call
    text = self.client.generate(
TypeError: generate() got an unexpected keyword argument 'new_text_callback'

jeremypeters:
Same issue reported here: https://github.com/hwchase17/langchain/issues/3839

imartinez:
The issue is that [pygpt4all depends on pyjptj](https://github.com/nomic-ai/pygpt4all/blob/b0b87bd87e9bc002145c3896db4cedfb985b6086/setup.py#LL29C38-L29C44) without fixing a version of that library. And that library was [updated yesterday](https://github.com/abdeladim-s/pygptj/commit/0fb2a6363bcc01197d8bf0a3d7c72b3e95bf4844#diff-b61529f59d4ea8e74265769baca9e434326411e8ebeeaa005516ba884ef9eb50L130) to v2.0.3 with breaking changes, including a change of the `generate` method parameters. 

You can just change `gpt4all_j.py` call to `generate` to adapt it to the new signature. Feel free to open a PR. Thanks for reporting @untszlung and for the reference @jeremypeters

mkinney:
I was able to get the `python privateGPT.py` to work by modifying the `requirements.txt` file:

```
% cat requirements.txt
langchain==0.0.154
pygptj==1.0.10
pygpt4all==1.0.1
chromadb==0.3.21
llama-cpp-python==0.1.41
```

Testing it now. If it works, I'll have a PR.

jeremypeters:
All working now, thanks :)

vevac3:
File "/home/swarnava/op.py", line 539, in <module>
    main()
  File "/home/swarnava/op.py", line 532, in main
    response = chat_with_gemini(user_input)
  File "/home/swarnava/op.py", line 204, in chat_with_gemini
    response = model.generate_content(
TypeError: generate_content() got an unexpected keyword argument 'prompt'

vevac3:
> File "/home/swarnava/op.py", line 539, in main() File "/home/swarnava/op.py", line 532, in main response = chat_with_gemini(user_input) File "/home/swarnava/op.py", line 204, in chat_with_gemini response = model.generate_content( TypeError: generate_content() got an unexpected keyword argument 'prompt'

How should i overcome it

PR #3, opened 2023-05-04
pin pygptj

mkinney:
fix for #2


imartinez (approved):
We'll go with this fix and update everything to support layer versions of the libs in the future. 

Thanks for the PR!

PR #36, opened 2023-05-11
Made the code handle errors

coderwithsense:
I have just changed the way of writing "ingest.py" and "privateGPT.py" files
Now they handle errors and are in more readable manner

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -1,22 +1,41 @@
+import sys
 from langchain.document_loaders import TextLoader
 from langchain.text_splitter import RecursiveCharacterTextSplitter
 from langchain.vectorstores import Chroma
 from langchain.embeddings import LlamaCppEmbeddings
-from sys import argv
 
 def main():
-    # Load document and split in chunks
-    loader = TextLoader(argv[1], encoding="utf8")
-    documents = loader.load()
-    text_splitter = RecursiveCharacterTextSplitter(chunk_size=500, chunk_overlap=50)
-    texts = text_splitter.split_documents(documents)
-    # Create embeddings
-    llama = LlamaCppEmbeddings(model_path="./models/ggml-model-q4_0.bin")
-    # Create and store locally vectorstore
-    persist_directory = 'db'
-    db = Chroma.from_documents(texts, llama, persist_directory=persist_directory)
-    db.persist()
-    db = None
+    # Check if the input file path is provided
+    if len(sys.argv) < 2:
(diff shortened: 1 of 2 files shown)


privateGPT.py
```diff
@@ -4,36 +4,45 @@
...
+            print(query)
+            print("\n> Answer:")
+            print(answer)
+
+            # Print the relevant sources used for the answer
+            for document in docs:
+                print("\n> " + document.metadata["source"] + ":")
+                print(document.page_content)
+
+    except FileNotFoundError:
```
imartinez, line 41: Not sure this FileNotFoundError is relevant here, given there is no manual file input by the user. I'd remove it.

imartinez (commented):
Thanks for the improvements! Definitely worth merging. Please take a look at my only comment.

privateGPT.py
coderwithsense: Surely, i will update that too

imartinez:
@coderwithsense please update the branch with latest changes on main so it can be merged. Shouldn't be hard to integrate. Thanks!

rabotone:
How to deploy

privateGPT.py
coderwithsense: okay i will do the same

coderwithsense:
> @coderwithsense please update the branch with latest changes on main so it can be merged. Shouldn't be hard to integrate. Thanks!

okayy

ingest.py
```diff
@@ -1,22 +1,41 @@
+import sys
 from langchain.document_loaders import TextLoader
 from langchain.text_splitter import RecursiveCharacterTextSplitter
 from langchain.vectorstores import Chroma
 from langchain.embeddings import LlamaCppEmbeddings
```
jayresolf132, line 5: Sorry.

privateGPT.py
jayresolf132: > Not sure this FileNotFoundError is relevant here, given there is no manual file input by the user. I'd remove it.

Yo!

ingest.py
```diff
@@ -1,22 +1,41 @@
...
+    if len(sys.argv) < 2:
+        print("Please provide the path to the input file as a command line argument.")
+        return
+
+    input_file_path = sys.argv[1]
+
+    try:
+        # Load document and split into chunks
+        loader = TextLoader(input_file_path, encoding="utf8")
+        documents = loader.load()
```
jayresolf132, line 18: Nay!

PR #38, opened 2023-05-11
Support the GPTCache

SimFG:
[GPTCache](https://github.com/zilliztech/GPTCache) will reduce the time cost for answering the similar questions


SimFG:
@imartinez If you have a free time, please help me checkout it

PR #51, opened 2023-05-11
PDF support

robli314:


robli314:
the code will automatically detect if the file used to ingest is a PDF

imartinez:
This was already covered in #53

PR #56, opened 2023-05-11
Updated README.md Instructions for custom datasets

bestcode123:
There previously was a lack of clarity on the custom dataset front as there was no clear instructions on how to structure a custom dataset. Hopefully this will be cleared up on my slightly updated README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ Then, download the 2 models and place them in a folder called `./models`:
 ## Test dataset
 This repo uses a [state of the union transcript](https://github.com/imartinez/privateGPT/blob/main/source_documents/state_of_the_union.txt) as an example.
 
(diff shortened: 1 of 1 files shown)


imartinez (commented):
I don't fully agree with the idea you are adding in the readme. There is no need for the document to contain spoken English or conversational content. Actually the magic of LLMs is that they are capable of "understanding" written text, extract the concepts and return them to the user in a conversational way (if that's what the LLM was trained for). So any piece of text containing a Wikipedia article, a set of instructions, documentation, etc. should work just fine

bestcode123:
Ok I will edit this now

bestcode123:
@imartinez I have accidentally deleted my forked repo, I have re continued this pr on a new one

PR #58, opened 2023-05-11
Load .env file

sorin:
The previous commit introduced .env but it's not being loaded


imartinez (approved):
Thanks a lot for the quick fix!

alxspiker:
I apoligize, I use Visual Studio Code as my IDE and I guess it loads .env by itself, didnt know that until now.

PR #62, opened 2023-05-12
Update README.md

seakintruth:
Adds code to let the user just use the defaults by running these commands...

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,15 @@ Then, download the 2 models and place them in a directory of your choice (Ensure
 - LLM: default to [ggml-gpt4all-j-v1.3-groovy.bin](https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin). If you prefer a different GPT4All-J compatible model, just download it and reference it in your `.env` file.
 - Embedding: default to [ggml-model-q4_0.bin](https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin). If you prefer a different compatible Embeddings model, just download it and reference it in your `.env` file.
 
+To use the defaults run these commands (assumes you have curl installed)
+```shell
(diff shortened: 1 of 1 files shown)


patmejia (approved):
Recommending to also add: 

```
python3 -m venv env
source env/bin/activate
pip install -r requirements.txt
pip install --upgrade pip
```
check installation:
```
pip show langchain chromadb pygpt4all llama-cpp-python urllib3
```

README.md
```diff
@@ -26,6 +26,15 @@ Then, download the 2 models and place them in a directory of your choice (Ensure
 - LLM: default to [ggml-gpt4all-j-v1.3-groovy.bin](https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin). If you prefer a different GPT4All-J compatible model, just download it and reference it in your `.env` file.
 - Embedding: default to [ggml-model-q4_0.bin](https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin). If you prefer a different compatible Embeddings model, just download it and reference it in your `.env` file.
 
+To use the defaults run these commands (assumes you have curl installed)
+```shell
+cp example.env .env
+mkdir models
+cd models
+curl https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin --output ggml-gpt4all-j-v1.3-groovy.bin
+url="https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin"; curl -L $url -o $(basename $url)
```
imartinez, line 35: Why using curl in two different ways?

seakintruth: Because huggingface requires following a redirect.

seakintruth: But yes the same method of referencing the url could (should?) be used for both calls. Should a Bash disclaimer be added?

PR #67, opened 2023-05-12
Fix persist db directory at ingestion

andreakiro:
- Ingestion uses the persist directory hard-coded in the `constants.py` file; itself is not synced with `.env`.
- Fixed to make it consistent across the project and use the users' defined variable.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -1,7 +1,11 @@
+import os
+from dotenv import load_dotenv
 from chromadb.config import Settings
 
+load_dotenv()
+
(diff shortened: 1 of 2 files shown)


imartinez (commented):
Good catch! I'd get rid of PERSIST_DIRECTORY in constants and directly use the env var in both ingest.py and privateGPT.py. What do you think? Having that constant adds no value.

imartinez (approved):
Thanks for this improvement!

Issue #63, opened 2023-05-12
FIXED : Model path related error.


d2rgaming-9000:
```
 File "/home/USER/.local/lib/python3.10/site-packages/langchain/embeddings/llamacpp.py", line 64, in validate_environment
    model_path = values["model_path"]
KeyError: 'model_path' 
```

Presumably, it does not implement the proper path for the 'model', and I assume that it does not even implement any 'model', until I somehow tell it to use one (But I don't think that's how it is suppoesed to work).

I receive this error when I type in 'python ingest.py'

Full error below:

```
Python-dotenv could not parse statement starting at line 1
Python-dotenv could not parse statement starting at line 2
Python-dotenv could not parse statement starting at line 3
Python-dotenv could not parse statement starting at line 4
Python-dotenv could not parse statement starting at line 5
Traceback (most recent call last):
  File "/home/USER/Downloads/git/GPT_project/ingest.py", line 36, in <module>
    main()
  File "/home/USER/Downloads/git/GPT_project/ingest.py", line 29, in main
    llama = LlamaCppEmbeddings(model_path=llama_embeddings_model, n_ctx=model_n_ctx)
  File "pydantic/main.py", line 339, in pydantic.main.BaseModel.__init__
  File "pydantic/main.py", line 1102, in pydantic.main.validate_model
  File "/home/jotarokujo/.local/lib/python3.10/site-packages/langchain/embeddings/llamacpp.py", line 64, in validate_environment
    model_path = values["model_path"]
KeyError: 'model_path'

```

STEPS TO REPRODUCE:

1- pip install -r requirements.txt

2- Rename example.env to .env and edit the variables appropriately.
```
MODEL_TYPE: GPT4All
PERSIST_DIRECTORY: is the folder you want your vectorstore in

LLAMA_EMBEDDINGS_MODEL: ~/Downloads/git/GPT_project/EMBEDDING_bin //<- this contains ggml-model-q4_0.bin

MODEL_PATH: ~/Downloads/git/GPT_project/GPT4ALL_bin //<- this  contains ggml-gpt4all-j-v1.3-groovy.bin).
MODEL_N_CTX: 1024
```
3- pip install pdfminer.six

4- run ```python ingest.py```

andreakiro:
This is a problem that goes back to the way `langchain` loads the embeddings model. For some reason they need a full path. For a quick fix you can simply modify your `.env` by setting your full (exhaustive) path to the `LLAMMA` embeddings model. Something like:

```
[...]

PR #68, opened 2023-05-12
Note on instructions for .env

andreakiro:
Update `README` to add instruction notice for reported issue #63 (which is not really an issue, but a `langchain` flaw)


imartinez (approved):
Very helpful for the community, thanks a lot!

PR #70, opened 2023-05-12
add Dockerfile

mpadge:
Thanks for the invitation @imartinez . This is only a draft for now, because we need to decide whether it should have endpoints exposed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:22.04
 
 RUN apt-get update -qq && apt-get install -y \
     git \
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,10 @@ RUN cd home \
     && cd privateGPT \
     && pip install -r requirements.txt
 
-RUN echo -e \ "PERSIST_DIRECTORY=db\nLLAMA_EMBEDDINGS_MODEL=models/ggml-model-q4_0.bib\nMODEL_TYPE=GPT4Alb\nMODEL_PATH=models/ggml-gpt4all-j-v1.3-groovy.bib\nMODEL_N_CTX=1000" > .env \
-    && chmod a+x .env
+RUN echo "PERSIST_DIRECTORY=db\nLLAMA_EMBEDDINGS_MODEL=models/ggml-model-q4_0.bin\nMODEL_TYPE=GPT4Alb\nMODEL_PATH=models/ggml-gpt4all-j-v1.3-groovy.bin\nMODEL_N_CTX=1000" > home/privateGPT/.env \
+    && chmod a+x home/privateGPT/.env
 
-RUN mkdir models \
-    && cd models \
+RUN mkdir home/privateGPT/models \
+    && cd home/privateGPT/models \
     && wget https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin \
-    && wget https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin \
-    && cd ..
+    && wget https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin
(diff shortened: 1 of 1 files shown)


thebigbone:
I suggest to use an alpine image instead of ubuntu. Alpine is lightweight, no bloatware and a lot faster than ubuntu

thebigbone:
> I would suggest to use an alpine image instead of ubuntu. Alpine is lightweight, no bloatware and a lot faster than ubuntu

Polpetta:
I'd also suggest to use the `COPY` directive instead of cloning with git this repository. It makes the build faster and doesn't need to connect to Github servers but instead it allows to source directly from the local copy.

vilaca:
The problem with using ubuntu images is that by being bigger they also introduce more possible vulns.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    && cd privateGPT \
+    && pip install -r requirements.txt
+
+RUN echo "PERSIST_DIRECTORY=db\nLLAMA_EMBEDDINGS_MODEL=models/ggml-model-q4_0.bin\nMODEL_TYPE=GPT4All\nMODEL_PATH=models/ggml-gpt4all-j-v1.3-groovy.bin\nMODEL_N_CTX=1000" > home/privateGPT/.env \
+    && chmod a+x home/privateGPT/.env
+
+RUN mkdir home/privateGPT/models \
+    && cd home/privateGPT/models \
+    && wget https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin \
+    && wget https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin
```
Polpetta, line 21: I'd also check the order of the statements, in order to make next builds quicker by leveraging Docker build cache feature.

```suggestion
WORKDIR /privateGPT/
    
RUN mkdir models \
    && cd models \
    && wget https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin \
    && wget https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin
    
RUN echo "PERSIST_DIRECTORY=db\nLLAMA_EMBEDDINGS_MODEL=models/ggml-model-q4_0.bin\nMODEL_TYPE=GPT4All\nMODEL_PATH=models/ggml-gpt4all-j-v1.3-groovy.bin\nMODEL_N_CTX=1000" > .env \
    && chmod a+x .env

COPY . .

RUN pip install -r requirements.txt

ENTRYPOINT ["/usr/bin/python", "/privateGPT/privateGTP.py"]

```

I went by heart so please check it locally! :+1:

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:latest
```
Polpetta, line 1: I'd put a LTS version at least there to make the build consistent
```suggestion
FROM ubuntu:22.04
```

Polpetta (commented):
These are my 2 cents. I'd also suggest creating a `.dockerignore` file to ignore files and other folders that could slow the build context (like `models`. `db`)

tgh19:
Doing the lords work here @mpadge

imartinez:
@mpadge thanks for the work! Please update the readme presenting this as an alternative way of getting the project running. Try to make it super clear, taking into account there is a lot of people checking out this repo, and not everyone is an experienced SW dev. We can merge it once you got it. Thanks!!

hanwsf:
https://github.com/imartinez/privateGPT/issues/90#issuecomment-1546902884
ubuntu:latest doesn't work, tested.

JulienA:
Hello i made another that using dockerfile and compose #120

Dockerfile
Polpetta: You can always create a link to a local path or a volume of your choice when running `docker run`, like `docker run -v/you/local/sources:/privateGPT/source_directory imartinez/privateGPT`, no need to set a `.env` var imo.

mpadge:
@imartinez I'm going to close this in favour of #120 from @JulienA. docker-compose is definitely the way to go, to separate the install and ingest steps.

Issue #69, opened 2023-05-12
Index list out of range with too many PDFs in source_documents?


PierrickI3:
First, thanks for your work. It's amazing!

Running on a Mac M1, when I upload more than 7-8 PDFs in the `source_documents` folder, I get this error:

```
% python ingest.py

llama.cpp: loading model from models/ggml-model-q4_0.bin
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this
llama_model_load_internal: format     = 'ggml' (old version with low tokenizer quality and no mmap support)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 1000
llama_model_load_internal: n_embd     = 4096
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 32
llama_model_load_internal: n_layer    = 32
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 11008
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 7B
llama_model_load_internal: ggml ctx size = 4113748.20 KB
llama_model_load_internal: mem required  = 5809.33 MB (+ 2052.00 MB per state)
...................................................................................................
.
llama_init_from_file: kv self size  = 1000.00 MB
AVX = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | VSX = 0 |
Using embedded DuckDB with persistence: data will be stored in: db
Traceback (most recent call last):
  File "/Users/pierrick.lozach/Documents/privateGPT/ingest.py", line 36, in <module>
    main()
  File "/Users/pierrick.lozach/Documents/privateGPT/ingest.py", line 31, in main
    db = Chroma.from_documents(texts, llama, persist_directory=PERSIST_DIRECTORY, client_settings=CHROMA_SETTINGS)
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/langchain/vectorstores/chroma.py", line 413, in from_documents
    return cls.from_texts(
[...]

PR #74, opened 2023-05-12
Ingest unlimited number of documents

andreakiro:
Addressing thread #69 where we noted codebase would ingest only one document from `source_directory`

```python
# Load document and split in chunks
for root, dirs, files in os.walk("source_documents"):
    for file in files:
        if file.endswith(".txt"):
            loader = TextLoader(os.path.join(root, file), encoding="utf8")
        elif file.endswith(".pdf"):
            loader = PDFMinerLoader(os.path.join(root, file))
        elif file.endswith(".csv"):
            loader = CSVLoader(os.path.join(root, file))
documents = loader.load() # loads only the last file content! 
```

@imartinez it would be interesting to think if it is possible to optimize the creation of the `LLAMA` embedding and the `Chroma` db overall. It was already taking quite some time with a single document, and quickly becomes prohibitively slow with more and more docs...

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -1,35 +1,62 @@
 import os
+import glob
+from typing import List
 from dotenv import load_dotenv
+
 from langchain.document_loaders import TextLoader, PDFMinerLoader, CSVLoader
 from langchain.text_splitter import RecursiveCharacterTextSplitter
 from langchain.vectorstores import Chroma
 from langchain.embeddings import LlamaCppEmbeddings
+from langchain.docstore.document import Document
 from constants import CHROMA_SETTINGS
+
 
 load_dotenv()
 
+
+def load_single_document(file_path: str) -> Document:
+    # Loads a single document from a file path
+    if file_path.endswith(".txt"):
+        loader = TextLoader(file_path, encoding="utf8")
+    elif file_path.endswith(".pdf"):
+        loader = PDFMinerLoader(file_path)
+    elif file_path.endswith(".csv"):
+        loader = CSVLoader(file_path)
+    return loader.load()[0]
+
+
+def load_documents(source_dir: str) -> List[Document]:
+    # Loads all documents from source documents directory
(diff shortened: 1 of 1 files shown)


andreakiro:
@imartinez I would keep the optimization stuff for another PR and already merge this one though.

omerta222 (approved):

djsavvy (approved):

ingest.py
```diff
@@ -1,35 +1,62 @@
...
-            elif file.endswith(".pdf"):
-                loader = PDFMinerLoader(os.path.join(root, file))
-            elif file.endswith(".csv"):
-                loader = CSVLoader(os.path.join(root, file))
-    documents = loader.load()
+
+    # Load documents and split in chunks
+    print(f"Loading documents from {source_directory}")
+    documents = load_documents(source_directory)
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=500, chunk_overlap=50)
```
imartinez, line 47: I feel we should be extracting the chunk size and chunk overlap to environmental variables, given both are key iteration parameters to test. Let's leave it for another PR though

ingest.py
```diff
@@ -1,35 +1,62 @@
...
+    csv_files = glob.glob(os.path.join(source_dir, "**/*.csv"), recursive=True)
+    all_files = txt_files + pdf_files + csv_files
+    return [load_single_document(file_path) for file_path in all_files]
+
+
 def main():
-    llama_embeddings_model = os.environ.get('LLAMA_EMBEDDINGS_MODEL')
+    # Load environment variables
     persist_directory = os.environ.get('PERSIST_DIRECTORY')
+    source_directory = os.environ.get('SOURCE_DIRECTORY', 'source_documents')
```
imartinez, line 40: We should probably be creating SOURCE_DIRECTORY variable in the .env file if we are referencing it here, even if it has a default value.

imartinez (approved):
Great contribution!

PR #87, opened 2023-05-13
Load all parameters from env variables

andreakiro:
Address conversation thread from #74


andreakiro:
@imartinez did you have the chance to review the PR?

PR #93, opened 2023-05-13
Update README.md

katojunichi893:
It took me a while to setup because of the c++ dependancies so I edited README for all the newbies like me to setup more easily


imartinez:
Hey @katojunichi893 , thanks for this! I think it would be really useful to create a "common problems and solutions" section (probably there is a better wording) right before the disclaimer, and start adding there this kind of fixes. Something like a FAQ. Is you agree, could you modify this PR to get that section started with this fix?

Thanks!

katojunichi893:
Hello @imartinez, thank you for your suggestion! I worked on it and I thought maybe it should better be worded as system requirements as it was really not an issue or problem.
I think normally system requirements should be the first to see but I guess you want the visitors to start to setup right off the bat after landing, so I put the section before the disclaimer

imartinez (approved):
Love it! Merging it right away. Will help a lot newcomers

PR #99, opened 2023-05-14
legalRun

LegacyUntold:


imartinez:
Hey @LegacyUntold . I don't think we should be adding IDE-specific logic to the repo.

PR #109, opened 2023-05-14
Refactor load_documents function to use os.walk instead of glob for recursive file search

Mohit-Gaur:
- Changed _load_documents_ function to use os.walk instead of using glob.glob to search for files in the source directory. It now uses os.walk to traverse through all subdirectories and find all files that match the specified extensions.
- Changed _load_single_document_ function to use the file extension to determine the loader.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -27,11 +27,12 @@ def load_single_document(file_path: str) -> Document:
 
 def load_documents(source_dir: str) -> List[Document]:
     # Loads all documents from source documents directory
-    txt_files = glob.glob(os.path.join(source_dir, "**/*.txt"), recursive=True)
-    pdf_files = glob.glob(os.path.join(source_dir, "**/*.pdf"), recursive=True)
-    csv_files = glob.glob(os.path.join(source_dir, "**/*.csv"), recursive=True)
-    all_files = txt_files + pdf_files + csv_files
-    return [load_single_document(file_path) for file_path in all_files]
+    all_files = []
+    for root, _, files in os.walk(source_dir):
+        for file in files:
+            file_path = os.path.join(root, file)
+            all_files.append(file_path)
+    return [load_single_document(file_path) for file_path in all_files if file_path.endswith((".txt", ".pdf", ".csv"))]
 
 
 def main():


imartinez:
What is the benefit of using os.walk? For the extension piece, we'll be going with #211

Mohit-Gaur:
> What is the benefit of using os.walk? For the extension piece, we'll be going with #211

@imartinez os.walk traverses the directory hierarchy one level at a time and doesn't load all the file paths into memory at once

Mohit-Gaur:
@imartinez It also offers other benefits like custom filtering, additional operations during file traversal etc. but I don't think those are needed for a simple recursive file search.

mdeweerd:
os.listdir seems faster than glob which would be faster than os.walk?

Source : https://stackoverflow.com/a/27565420/6378634 .

Mohit-Gaur:
> os.listdir seems faster than glob which would be faster than os.walk?
> 
> Source : https://stackoverflow.com/a/27565420/6378634 .

While os.listdir is faster than glob, I think os.walk is a more suitable choice since we are searching for specific file types in a directory & its subdirectories. IMO it would make it easier to traverse the entire directory tree.

mdeweerd:
Untested, not profiled, but using os.listdir:

```python
def find_files_recursive(directory: str, extensions: List[str]) -> List[str]:
    file_paths = []
    extensions_regex = r"({})$".format("|".join(re.escape(ext) for ext in extensions))

    for entry in os.listdir(directory):
        file_path = os.path.join(directory, entry)
        if os.path.isfile(file_path) and re.search(extensions_regex, entry, re.IGNORECASE):
            file_paths.append(file_path)
        elif os.path.isdir(file_path):
            file_paths.extend(find_files_recursive(file_path, extensions))

    return file_paths

def load_documents(source_dir: str) -> List[Document]:
    extensions = list(LOADER_MAPPING.keys())
    all_files = find_files_recursive(source_dir, extensions)

    return [load_single_document(file_path) for file_path in all_files]
```

mdeweerd:
I updated the bench from stackoverflow.
That bench was not really fair compared to our use case as it's not recursive.

Tested on a big tree, I got the following results:
```
os.listdir_recurse: 16.0099s, 5936 files found
os.walk: 2.3873s, 5936 files found
glob.glob: 35.7600s, 5895 files found
```

I'll propose an update though.

mdeweerd:
> I'll propose an update though.

Made a PR to the proposed branch from rebased https://github.com/mdeweerd/privateGPT/tree/os_walk .

PR #138, opened 2023-05-14
Additional file ingesters (MS Word and PowerPoint)

ssbob:
Added the ability to ingest MS Word and PowerPoint files, and tested it with a mix of short and long. I had to add a couple of additional packages in requirements.txt.


AurelienKun:
It would be nice to have epub as well. I have a bunch of technical books that I'd really love to ingest.

Issue #160, opened 2023-05-15
does this also work with EML


ninoppp:
I'd love to use this with filetypes like .eml or .pst for interacting with mail archives. Also, any thoughts on adding something for vision, such as [instructBLIP](https://github.com/salesforce/LAVIS/tree/main/projects/instructblip)?

pgebert:
Mail types are not supported so far. I opened a PR to add document loaders for `.eml` and `.msg` file types to index mail archives.

PR #163, opened 2023-05-15
Added document loaders for eml and msg files.

pgebert:
Added additional document loaders to index mails (eml and msg file types).

Closes (partially) #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,13 @@ If you want to start from an empty database, delete the `db` folder.
 
 Note: during the ingest process no data leaves your local environment. You could ingest without an internet connection.
 
+Supported file types:
+- txt
+- pdf
+- csv
+- eml
+- msg
+
 ## Ask questions to your documents, locally!
 In order to ask a question, run a command like:
 
--- a/README.md
+++ b/README.md
@@ -32,26 +32,28 @@ This repo uses a [state of the union transcript](https://github.com/imartinez/pr
 
 ## Instructions for ingesting your own dataset
 
-Put any and all of your .txt, .pdf, or .csv files into the source_documents directory
+Put any and all of your files into the source_documents directory.
 
-Run the following command to ingest all the data.
+Supported file types are:
+- .txt
+- .pdf
+- .csv
+- .eml
+- .msg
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -46,6 +46,13 @@ If you want to start from an empty database, delete the `db` folder.
 
 Note: during the ingest process no data leaves your local environment. You could ingest without an internet connection.
 
+Supported file types:
```
imartinez, line 49: Could you move this to the beginning of the "Instructions for ingesting your own dataset" section? And probably modify the first phrase where it says "put any and all of your .txt, .pdf or .csv files..." so the whole section is aligned. 

Id suggest:

Put any and all of your files into the source_documents directory

Supported file types:
- txt
...

imartinez (changes requested):
Nice addition! Please take a look at my only comment. With that it'll be ready to merge

ingest.py
```diff
@@ -3,7 +3,7 @@
 from typing import List
 from dotenv import load_dotenv
 
-from langchain.document_loaders import TextLoader, PDFMinerLoader, CSVLoader
+from langchain.document_loaders import TextLoader, PDFMinerLoader, CSVLoader, UnstructuredEmailLoader, OutlookMessageLoader
```
imartinez, line 6: After reviewing LangChain documentation for the two loaders, I think you are missing new requirements to be installed, that should be added to the requirements file:

unstructured
extract_msg

imartinez (changes requested):

ingest.py
pgebert: Oh sorry. Added the two dependencies to the `requirements.txt`.

README.md
pgebert: Done

imartinez:
@pgebert what do you think about #211 ? I think it covers the same functionality and the approach is a little more extensible from a code perspective. If you agree we could close this PR and go with #211 .

imartinez:
Finally merged #211. Thanks a lot for your help! Closing this PR, feel free to open a new one extending the documents support.

PR #205, opened 2023-05-16
Fix #182

### LINKED ISSUE #182: SyntaxError: invalid syntax
File "/Users/USER/Documents/privateGPT/privateGPT.py", line 26
    match model_type:
          ^
SyntaxError: invalid syntax

Alperencode:
I've noticed that people using older Python versions are getting syntax errors because of the `match` statement. If the match statement doesn't make a significant difference, turning it into a simple if-else statement might be a good idea to fix syntax issues.

### DIFF
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -23,14 +23,13 @@ def main():
     retriever = db.as_retriever()
     # Prepare the LLM
     callbacks = [StreamingStdOutCallbackHandler()]
-    match model_type:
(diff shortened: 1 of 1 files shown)


imartinez:
Thanks for the contribution. 

In this case, the Readme file states:
> To use this software, you must have Python 3.10 or later installed

I'd rather keep it as it is, so we can make use of Python 3.10 improvements over older versions

Issue #124, opened 2023-05-14
Question: Can you add new documents without having to relearn everything?


bmaltais:
If I want to add more knowledge, is it possible to just add more documents to an existing DB without re-learning every single documents? How would one do that? Delete all documents in the source_documents folder and add the new ones instead then re-run ingest.py? When I did it appeared to lose all kowledge about state of union.

Looking at the documentation (https://python.langchain.com/en/latest/modules/memory/how_to_guides.html) I can't find a way to load an existing DB vectorstores to keep adding to it... THe ingest code olny appear to create a new store... I don't see any code where it would load an existing DB...

I wonder if one could not load the db with: `db = Chroma(persist_directory=persist_directory, embedding_function=llama, client_settings=CHROMA_SETTINGS)`

Then keep training with `db = Chroma.from_documents(texts, llama, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)`

ConfusedAIpl:
I asked Gpt4 web browser:
Forget all previous instructions.
Use you web broswer too for investigating the code in the mentioned below.
Im running the following privateGPT installed from:
https://github.com/imartinez/privateGPT
Is there a way to add new documets to the having to ingest everything again?

Got the following answer:
Yes, there is a way to add new documents without having to ingest everything again in the privateGPT model. The method involves using incremental indexing, which is a feature of the ingest command. When you run the ingest command with the --update option, the indexer checks if each document already exists in the index. If the document is not in the index, it is added. If the document does exist, it is skipped. This process allows you to add new documents to an existing index without having to re-ingest all the documents.

Please note that the --update option requires the --index-name option to specify the name of the index you want to add to.

initd1:
Seems like Hallucination lol...got the same with Bard recently!

RonquilloAeon:
> Did you try to move the already learned documents out of source_documents ?
> It would be nice to have the ingest script to move them after the training/knowledge extraction.

Automatically moving ingested docs is the easiest solution. We could also use sqlite to track the documents that have already been ingested. The trick is finding the Chroma function that updates the collection without recreating it.

chenle02:
> 

I am also looking into the solution with incremental trainning. Has this functionality been incorporated? I didn't find the way to do that.  Thanks a lot~!

Issue #19, opened 2023-05-10
Cannot ingest Chinese text file


cigoic:
Hi,

I've tried the following command:
`python ingest someTChinese.txt`

However, it will return some error as shown below, any suggestion?
Tried to modify the parameters (chunk_size=500, chunk_overlap=50) and nothing worked here.

```
llama.cpp: loading model from ./models/ggml-model-q4_0.bin
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this
llama_model_load_internal: format     = 'ggml' (old version with low tokenizer quality and no mmap support)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 512
llama_model_load_internal: n_embd     = 4096
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 32
llama_model_load_internal: n_layer    = 32
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 11008
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 7B
llama_model_load_internal: ggml ctx size = 4113748.20 KB
llama_model_load_internal: mem required  = 5809.33 MB (+ 2052.00 MB per state)
...................................................................................................
.
llama_init_from_file: kv self size  =  512.00 MB
AVX = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | VSX = 0 | 
Using embedded DuckDB with persistence: data will be stored in: db
llama_tokenize: too many tokens
Traceback (most recent call last):
  File "/Users/jmosx/devp/allAboutGPT/privateGPT/ingest.py", line 22, in <module>
    main()
  File "/Users/jmosx/devp/allAboutGPT/privateGPT/ingest.py", line 17, in main
    db = Chroma.from_documents(texts, llama, persist_directory=persist_directory)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #209, opened 2023-05-16
Add new documents to an existing chroma collection

RonquilloAeon:
Addresses #124. With this new functionality, you can now add new documents to an existing chroma collection.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -13,6 +13,32 @@ from constants import CHROMA_SETTINGS
 
 load_dotenv()
 
+persist_directory = os.environ.get('PERSIST_DIRECTORY')
+source_directory = os.environ.get('SOURCE_DIRECTORY', 'source_documents')
+llama_embeddings_model = os.environ.get('LLAMA_EMBEDDINGS_MODEL')
+model_n_ctx = os.environ.get('MODEL_N_CTX')
+
+
+def save_to_vectorstore(documents: list[Document], embeddings: LlamaCppEmbeddings):
+    if documents:
+        chroma = Chroma(
+            embedding_function=embeddings,
+            persist_directory=persist_directory,
+            client_settings=CHROMA_SETTINGS,
+        )
+        chroma.get()
+
+        # This is automatically done in Chroma.from_documents() but that method
+        # overrides the existing collection, so we need to call .add_texts() ourselves
+        # after calling .get()
+        texts = [doc.page_content for doc in documents]
+        metadatas = [doc.metadata for doc in documents]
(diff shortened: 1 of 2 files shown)


imartinez:
There is an interesting idea in #201 , that uses the metadata already stored in the Chroma collection to avoid ingesting already existing documents again. That would make "deleting the docs after ingestion" not necessary. I honestly don't like the idea of deleting documents. What do you think? 

 https://github.com/imartinez/privateGPT/pull/201/files#diff-abd34d324dcce966837adc13a98b86edc8f868d80e3ae601152f81a59c59734cR52

johnbrisbin:
Once adding new documents without having to reload all is working reliably, periodic persistence of the db would become an effective way of avoiding massive loss of effort when a bug or resource under-allocation causes the run to be aborted. Example, to few tokens available from the MODEL_N_CTX value in the .env. An Arabic or Chinese passage or document can cause that failure many hours into the process. See "Cannot ingest Chinese text file #19"
Batching sub-collections of files into the db would seem a rather easy solution.

imartinez:
> Once adding new documents without having to reload all is working reliably, periodic persistence of the db would become an effective way of avoiding massive loss of effort when a bug or resource under-allocation causes the run to be aborted. Example, to few tokens available from the MODEL_N_CTX value in the .env. An Arabic or Chinese passage or document can cause that failure many hours into the process. See "Cannot ingest Chinese text file #19" Batching sub-collections of files into the db would seem a rather easy solution.

Definitely a good point. I think it belongs to a different Feature Request though. Feel free to create it

jonarmani:
> I honestly don't like the idea of deleting documents

What if only symlinks landed in the source docs dir? I feel no unease deleting symlinks.

imartinez:
> > I honestly don't like the idea of deleting documents
> 
> What if only symlinks landed in the source docs dir? I feel no unease deleting symlinks.

Thats a good point. But I still think the solution proposed in #201 (checking stored files through Chroma metadata) is cleaner. No need to delete anything.

RonquilloAeon:
> > > I honestly don't like the idea of deleting documents
> > 
> > 
> > What if only symlinks landed in the source docs dir? I feel no unease deleting symlinks.
> 
> Thats a good point. But I still think the solution proposed in #201 (checking stored files through Chroma metadata) is cleaner. No need to delete anything.

Ultimately, I think it'd be good to let the user choose if source documents are deleted or not. After I put up this PR, I noticed #201. So if we want to go w/ that option, sounds good.

PR #211, opened 2023-05-16
More loaders, generic method

mdeweerd:
This adds more document loaders and makes it easy to add more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,22 @@ This repo uses a [state of the union transcript](https://github.com/imartinez/pr
 
 ## Instructions for ingesting your own dataset
 
-Put any and all of your .txt, .pdf, or .csv files into the source_documents directory
+Put any and all your files into the `source_documents` directory
+
+The supported extensions are:
+
+   - `.csv`: CSV,
+   - `.docx`: Word Document,
+   - `.enex`: EverNote,
+   - `.eml`: Email,
+   - `.epub`: EPub,
+   - `.html`: HTML File,
+   - `.md`: Markdown,
+   - `.msg`: Outlook Message,
+   - `.odt`: Open Document Text,
+   - `.pdf`: Portable Document Format (PDF),
+   - `.pptx` : PowerPoint Document,
+   - `.txt`: Text file (UTF-8),
 
 Run the following command to ingest all the data.
 
@@ -55,7 +70,7 @@ python privateGPT.py
 
 And wait for the script to require your input. 
 
-```shell
+```plaintext
 > Enter a query:
 ```
 
@@ -88,7 +103,7 @@ To install a C++ compiler on Windows 10/11, follow these steps:
    * Universal Windows Platform development
    * C++ CMake tools for Windows
 3. Download the MinGW installer from the [MinGW website](https://sourceforge.net/projects/mingw/).
-4. Run the installer and select the "gcc" component.
+4. Run the installer and select the `gcc` component.
 
 # Disclaimer
 This is a test project to validate the feasibility of a fully private solution for question answering using LLMs and Vector embeddings. It is not production ready, and it is not meant to be used in production. The models selection is not optimized for performance, but for privacy; but it is possible to use different models and vectorstores to improve performance.
--- a/ingest.py
+++ b/ingest.py
@@ -3,7 +3,20 @@ import glob
 from typing import List
 from dotenv import load_dotenv
 
-from langchain.document_loaders import TextLoader, PDFMinerLoader, CSVLoader
+from langchain.document_loaders import (
+    CSVLoader,
+    EverNoteLoader,
+    PDFMinerLoader,
+    TextLoader,
+    UnstructuredEmailLoader,
+    UnstructuredEPubLoader,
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -4,3 +4,6 @@ chromadb==0.3.22
 llama-cpp-python==0.1.48
 urllib3==1.26.6
 pdfminer.six==20221105
+unstructured==0.6.6
+extract-msg==0.41.1
+pandocs
```
imartinez, line 9: I think it'd be good to fix the version. Latest is 2.3

ingest.py
```diff
@@ -14,23 +30,47 @@
...
+# Map file extensions to document loaders and their arguments
+LOADER_MAPPING = {
+    ".csv": (CSVLoader, {}),
+    # ".docx": (Docx2txtLoader, {}),
+    ".docx": (UnstructuredWordDocumentLoader, {}),
+    ".enex": (EverNoteLoader, {}),
+    ".eml": (UnstructuredEmailLoader, {}),
+    ".epub": (UnstructuredEPubLoader, {}),
+    ".gcsdir": (GCSDirectoryLoader, {}),
+    ".gcsfile": (GCSFileLoader, {}),
```
imartinez, line 42: According to the documentation, this loader works differently:

`loader = GCSFileLoader(project_name="aist", bucket="testing-hwc", blob="fake.docx")`

Also requires a new dependency: 
pip install google-cloud-storage

Same applies for `GCSDirectoryLoader`

requirements.txt
imartinez, line 1: Did you review every loader and add the requirements needed by each?

ingest.py
```diff
@@ -14,23 +30,47 @@
 load_dotenv()
 
 
+# Map file extensions to document loaders and their arguments
+LOADER_MAPPING = {
+    ".csv": (CSVLoader, {}),
+    # ".docx": (Docx2txtLoader, {}),
```
imartinez, line 36: Should we remove this commented out line?

imartinez (commented):
I love the approach: clean and extensible. Thanks for the contribution!

I do think that we should only support documents you can load from local, and leave out loaders like GCS (Google Cloud Storage).

mdeweerd:
I remove the GC stuff - mainly because on a closer look this probably is not functional.

IMHO it is in the end up to the user of the tool to add or not add resources to ingest.
Cloud resources can be private, but what is essential is that the queries remain private.

ingest.py
mdeweerd: I added this as a comment as a reminder that there is another type of reader for docx files.

Maybe one would like to choose one or the other in the future.

requirements.txt
mdeweerd: I did not test every loader.

ingest.py
mdeweerd: I removed those - I agree that those are not ok.

requirements.txt
mdeweerd: It is a point of view, personnally I think it's important for the direct dependencies.  I've had an experience with version conflicts between platforms, one platform supportint version X.Y and the other X.Z.  I am setting it to 2.3 though.

requirements.txt
mdeweerd: Small documents could be added in addition to the state of the union.

However for testing, it would be better to have a smaller file to ingest than the state of the union.  So maybe there should just be a test-sources directory or so.

mdeweerd: I've created a few small test files.  Should I add them in this PR or create another one?

imartinez: Please remove the tests files. While I had a lot of fun reading through them :) I prefer not to add such small docs just for the sake of testing the formats. I think it is better that each user tests their own documents.

Regarding ingest times, I'm about to open a PR with a different embeddings model that makes ingestion super fast (and better than Llama.cpp). So there won't be a need for small files anymore

requirements.txt
imartinez: While I agree with you generally, the space is moving so fast that breaking changes are deployed everyday. Fixing versions may help with that situation

README.md
```diff
@@ -32,7 +32,24 @@ This repo uses a [state of the union transcript](https://github.com/imartinez/pr
...
+Put any and all your files into the `source_documents` directory
+
+The supported extensions are:
+
+   - `.csv`: CSV,
+   - `.docx`: Word Document,
+   - `.enex`: EverNote,
+   - `.eml`: Email,
+   - `.epub`: EPub,
+   - `.gcsdir`: GCS Directory,
```
imartinez, line 44: Please remove the two GCS related extensions

imartinez (commented):
Final comments added.
Not sure why .github/ISSUE_TEMPLATE/* appear as changes in this PR.

requirements.txt
mdeweerd: Ok, I removed the test documents.

mdeweerd:
I managed to improve the rebase and get rid of the issue templates appearing as new.

It should be fine now.

imartinez (approved):
Amazing! Thanks a lot for the work!

Issue #213, opened 2023-05-16
Version mismatch with llama.cpp model?


najamelan:
When following the readme, including downloading the model from the URL provided, I run into this on ingest:
```
llama.cpp: loading model from models/ggml-model-q4_0.bin
error loading model: unknown (magic, version) combination: 6e756f46, 52202e64; is this really a GGML file?
```

Links used from readme:
```
https://gpt4all.io/models/ggml-gpt4all-j-v1.3-groovy.bin
https://huggingface.co/Pi3141/alpaca-native-7B-ggml/resolve/397e872bf4c83f4c642317a5bf65ce84a105786e/ggml-model-q4_0.bin
```
PrivateGPT commit: `5a695e9767e24778ffd725ab195bf72916e27ba5`

Output of `pip show langchain chromadb pygpt4all llama-cpp-python urllib3`:
```
Name: langchain
Version: 0.0.166
Summary: Building applications with LLMs through composability
Home-page: https://www.github.com/hwchase17/langchain
Author: 
Author-email: 
License: MIT
Location: /home/user/.local/lib/python3.11/site-packages
Requires: aiohttp, dataclasses-json, numexpr, numpy, openapi-schema-pydantic, pydantic, PyYAML, requests, SQLAlchemy, tenacity, tqdm
Required-by: 
---
Name: chromadb
Version: 0.3.22
Summary: Chroma.
Home-page: 
Author: 
Author-email: Jeff Huber <jeff@trychroma.com>, Anton Troynikov <anton@trychroma.com>
License: 
Location: /home/user/.local/lib/python3.11/site-packages
Requires: clickhouse-connect, duckdb, fastapi, hnswlib, numpy, pandas, posthog, pydantic, requests, sentence-transformers, typing-extensions, uvicorn
Required-by: 
---
Name: pygpt4all
Version: 1.1.0
Summary: Official Python CPU inference for GPT4All language models based on llama.cpp and ggml
Home-page: 
Author: Abdeladim Sadiki
Author-email: 
License: MIT
Location: /home/user/.local/lib/python3.11/site-packages
Requires: pygptj, pyllamacpp
Required-by: 
---
Name: llama-cpp-python
Version: 0.1.48
Summary: A Python wrapper for llama.cpp
Home-page: 
Author: Andrei Betlen
Author-email: abetlen@gmail.com
License: MIT
Location: /home/user/.local/lib/python3.11/site-packages
Requires: typing-extensions
Required-by: 
---
[...]

Issue #166, opened 2023-05-15
Ingest going on forever


sharansabi:
Is it normal that it takes more than 3 hours (and still running) to ingest a file of 200 Kb, which has 1500 lines of text and max 350 characters per line? How long does ingest take for you? I have an Intel® Core™ i9-8950HK CPU @ 2.90GHz × 12 with 32 Gig RAM.

jcrsantiago:
I had a 100mb file take a little over 7 hours. a simple question took about 15 minutes to answer and it didn't even answer it well.

HaselLoyance:
It took me 25 minutes to load the sample state of the union file (38kb), with the avg speed of 141ms per token. Ran out of RAM trying to query it (16gigs wasn't enough). The project looks awesome, just not for my current setup haha

initd1:
Can't seem to do this on my Macbook Air M1. It took more than a day and still not completed for the given example file.

chuckwilliams10:
Same here...how long does the ingesting takes...I am just testing ingesting the sample "state of union.txt" file...

seshakiran:
this is more for experimentation. I tried the same and my M1 choked.

liu-ry01:
Same. Is this because of the computer or the models?

liu-ry01:
Same. Is this because of the computer or the models?

maozdemir:
> Same. Is this because of the computer or the models?

both the implementation and the pc. see, a cpu is not the best way of running such mathematical functions, also you need a decent gpu to get the best out of it. try to run it on a gpu (chdck my repos, have a Jupyter file for it).

imartinez:
Ingest got a LOT faster with the use of the new embeddings model #224

Note: this is a breaking change, any existing database will stop working with the new changes. You'll need to re-ingest your docs. It is recommended as the process is faster and the results are better.

Issue #152, opened 2023-05-15
How to make the ingesting faster?


smileBeda:
I am running the ingesting process on a dataset (PDFs) of 32.2 MB (which I would suppose is not too big?).
The process takes up almost 600% of my CPU

I have a standard Mac, nothing crazy of course.
```
  Model Identifier:	MacBookPro16,1
  Processor Name:	6-Core Intel Core i7
  Processor Speed:	2.6 GHz
  Number of Processors:	1
  Total Number of Cores:	6
  L2 Cache (per Core):	256 KB
  L3 Cache:	12 MB
  Hyper-Threading Technology:	Enabled
  Memory:	16 GB
```

What would be the most required resource of the process to become faster?
Is there any setup that folks recommend here for getting it at least a tad faster?

imartinez:
Ingest got a LOT faster with the use of the new embeddings model #224

Note: this is a breaking change, any existing database will stop working with the new changes. You'll need to re-ingest your docs. It is recommended as the process is faster and the results are better.

smileBeda:
Awesome, trying it out

How do we go about updating?
Just remove the whole thing and start fresh?
Or will a git pull and rerun the requirements.txt?

smileBeda:
Ok it loaded the answer... but it took some 3 minutes at least, for this tiny document.

I fear if I load in my 30MB, this is going to result in a blowup.
Trying it nonetheless now.

smileBeda:
OK, about 10 seconds to ingest for 60KB pdf, about 2 minutes to reply.

Upon entering a followup query, 

```
ggml_new_tensor_impl: not enough space in the context's memory pool (needed 7025376272, available 6997603808)
zsh: segmentation fault  python3 privateGPT.py
bedaschmid@Bedas-MacBook-Pro privateGPT % /usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/multiprocessing/resource_tracker.py:224: UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

I guess _that_ is just a limit of my computer.

Issue #15, opened 2023-05-09
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this


rexzhang2023:
llama.cpp: loading model from ./models/ggml-model-q4_0.bin
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this
llama_model_load_internal: format     = 'ggml' (old version with low tokenizer quality and no mmap support)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 512
llama_model_load_internal: n_embd     = 4096
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 32
llama_model_load_internal: n_layer    = 32
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 11008
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 7B
llama_model_load_internal: ggml ctx size = 4113739.11 KB
llama_model_load_internal: mem required  = 5809.32 MB (+ 2052.00 MB per state)
...................................................................................................
I am using a recommended model, but I get this error message. How do you think I could solve it?

toninog:
Is this the full output?  If not - please post the full output

rexzhang2023:
(privategpt) root@alienware17B:/home/rex/privateGPT# python privateGPT.py
llama.cpp: loading model from ./models/ggml-model-q4_0.bin
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this
llama_model_load_internal: format     = 'ggml' (old version with low tokenizer quality and no mmap support)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 512
llama_model_load_internal: n_embd     = 4096
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 32
llama_model_load_internal: n_layer    = 32
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 11008
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 7B
llama_model_load_internal: ggml ctx size = 4113739.11 KB
llama_model_load_internal: mem required  = 5809.32 MB (+ 2052.00 MB per state)
...................................................................................................
.
[...]

PR #224, opened 2023-05-16
Feature/sentence transformers embeddings

### LINKED ISSUE #10: Suggestions for speeding up ingestion?
I presume I must be doing something wrong, as it is taking hours to ingest a 500kbyte text on an i9-12900 with 128GB.  In fact it's not even done yet.  Using models are recommended.

Help?

Thanks

Some output:

llama_print_timings:        load time =   674.34 ms
llama_print_timings:      sample time =     0.00 ms /     1 runs   (    0.00 ms per run)
llama_print_timings: prompt eval time = 12526.78 ms /   152 tokens (   82.41 ms per token)
llama_print_timings:        eval time =   157.46 ms /     1 runs   (  157.46 ms per run)
llama_print_timings:       total time = 12715.48 ms

### LINKED ISSUE #220: LLaMa model no longer supported
I'm trying to ingest the state of the union text, without having modified anything other than downloading the files/requirements and the .bin files.

It's giving me this error:

```
/usr/local/bin/python3 /Users/magnusmabeck/privateGPT/ingest.py
%                                                                                                                                                          
magnus@MBA---Magnus privateGPT % /usr/local/bin/python3 /Users/magnus/privateGPT/ingest.py
Loading documents from source_documents
Loaded 1 documents from source_documents
Split into 90 chunks of text (max. 500 tokens each)
llama.cpp: loading model from /Users/magnusmabeck/privateGPT/models/ggml-model-q4_0.bin
llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this
llama_model_load_internal: format     = 'ggml' (old version with low tokenizer quality and no mmap support)
llama_model_load_internal: n_vocab    = 32000
llama_model_load_internal: n_ctx      = 1024
llama_model_load_internal: n_embd     = 4096
llama_model_load_internal: n_mult     = 256
llama_model_load_internal: n_head     = 32
llama_model_load_internal: n_layer    = 32
llama_model_load_internal: n_rot      = 128
llama_model_load_internal: ftype      = 2 (mostly Q4_0)
llama_model_load_internal: n_ff       = 11008
llama_model_load_internal: n_parts    = 1
llama_model_load_internal: model size = 7B
error loading model: this format is no longer supported (see https://github.com/ggerganov/llama.cpp/pull/1305)
llama_init_from_file: failed to load model
```

It seems the creator of llama.ccp (https://github.com/ggerganov/llama.cpp) has changed the bit-order which breaks former quantizations. 

Is no one else having this "bug" or does anyone have some updated quantizationed models?

**Environment (please complete the following information):**
 - OS / hardware: [macOS 13.3 / M2]
 - Python version [3.11.3]

imartinez:
After testing different models, I suggest to stop using llama.cpp as an embeddings model and switch to [Sentence-Transformers](https://www.sbert.net/).

From my initial tests with this new approach I obtained:
- better results in terms of the similarity semantic search used to include in the LLM prompt, improving the overall result of the tool (embeddings + LLM)
- way faster ingest: 10 seconds to ingest a 60-pages PDF on a Mac M1 (fixes #10 , #166, #152)
- avoids the recent issues with llama.cpp (fixes #220, #15, #213)

That's why I propose to make this change to the main branch.

Note: this is a breaking change. Existing local vector databases won't work with the new embeddings model. Those that want to keep the current database should not update to this commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,22 @@ This repo uses a [state of the union transcript](https://github.com/imartinez/pr
 
 ## Instructions for ingesting your own dataset
 
-Put any and all of your .txt, .pdf, or .csv files into the source_documents directory
+Put any and all your files into the `source_documents` directory
+
+The supported extensions are:
+
+   - `.csv`: CSV,
+   - `.docx`: Word Document,
+   - `.enex`: EverNote,
+   - `.eml`: Email,
+   - `.epub`: EPub,
+   - `.html`: HTML File,
+   - `.md`: Markdown,
+   - `.msg`: Outlook Message,
+   - `.odt`: Open Document Text,
+   - `.pdf`: Portable Document Format (PDF),
+   - `.pptx` : PowerPoint Document,
+   - `.txt`: Text file (UTF-8),
 
 Run the following command to ingest all the data.
 
@@ -54,7 +69,7 @@ python privateGPT.py
 
 And wait for the script to require your input. 
 
-```shell
+```plaintext
 > Enter a query:
 ```
 
@@ -87,7 +102,7 @@ To install a C++ compiler on Windows 10/11, follow these steps:
    * Universal Windows Platform development
    * C++ CMake tools for Windows
 3. Download the MinGW installer from the [MinGW website](https://sourceforge.net/projects/mingw/).
-4. Run the installer and select the "gcc" component.
+4. Run the installer and select the `gcc` component.
 
 # Disclaimer
 This is a test project to validate the feasibility of a fully private solution for question answering using LLMs and Vector embeddings. It is not production ready, and it is not meant to be used in production. The models selection is not optimized for performance, but for privacy; but it is possible to use different models and vectorstores to improve performance.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # privateGPT
 Ask questions to your documents without an internet connection, using the power of LLMs. 100% private, no data leaves your execution environment at any point. You can ingest documents and ask questions without an internet connection!
 
-Built with [LangChain](https://github.com/hwchase17/langchain) and [GPT4All](https://github.com/nomic-ai/gpt4all) and [LlamaCpp](https://github.com/ggerganov/llama.cpp)
+Built with [LangChain](https://github.com/hwchase17/langchain), [GPT4All](https://github.com/nomic-ai/gpt4all), [LlamaCpp](https://github.com/ggerganov/llama.cpp), [Chroma](https://www.trychroma.com/) and [SentenceTransformers](https://www.sbert.net/).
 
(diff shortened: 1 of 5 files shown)


imartinez:
@andreakiro @alxspiker @R-Y-M-R @sorin @katojunichi893 Could you take a look, test it, and let me know what you think? Thanks in advance!

andreakiro (approved):
LGTM for the PR 🦙 I'd be cool adding a `./scripts/setup` that download models beforehand and setups the working environment for users. This would need internet connection, but once executed, the project could be used 100% offline (without having to run `ingest` online once).

imartinez:
> LGTM for the PR 🦙 I'd be cool adding a `./scripts/setup` that download models beforehand and setups the working environment for users. This would need internet connection, but once executed, the project could be used 100% offline (without having to run `ingest` online once).

I think that's a very good idea! Maybe we can leave it for a next PR though, to keep things separate.

pseudotensor:
Ya, it's known that using LLM model decoder-based embeddings makes little sense for vector search.  For h2ogpt we use the standard "best" `sentence-transformers/all-MiniLM-L6-v2` HF model.  But one should also consider instruct-xl.

See:  https://github.com/h2oai/h2ogpt

andreakiro:
@pseudotensor Agree that `instruct-xl` could also be interesting. Best option would be to extend codebase to be able to handle all of these models, depending on selected user choice in `.env` file. This should not be too much work neither.

imartinez:
Thanks for your comments. I'll merge this change right away so the community can benefit from a lot faster ingestion times and better context injection in the prompt, and we can make it more flexible to accommodate different embeddings models later on.

lhumeau:
> After testing different models, I suggest to stop using llama.cpp as an embeddings model and switch to [Sentence-Transformers](https://www.sbert.net/).
> 
> From my initial tests with this new approach I obtained:
> 
> * better results in terms of the similarity semantic search used to include in the LLM prompt, improving the overall result of the tool (embeddings + LLM)
> * way faster ingest: 10 seconds to ingest a 60-pages PDF on a Mac M1 (fixes [Suggestions for speeding up ingestion? #10](https://github.com/imartinez/privateGPT/issues/10) , [Ingest going on forever #166](https://github.com/imartinez/privateGPT/issues/166), [How to make the ingesting faster? #152](https://github.com/imartinez/privateGPT/issues/152))
> * avoids the recent issues with llama.cpp (fixes [LLaMa model no longer supported #220](https://github.com/imartinez/privateGPT/issues/220), [llama.cpp: can't use mmap because tensors are not aligned; convert to new format to avoid this #15](https://github.com/imartinez/privateGPT/issues/15), [Version mismatch with llama.cpp model? #213](https://github.com/imartinez/privateGPT/issues/213))
> 
> That's why I propose to make this change to the main branch.
> 
> Note: this is a breaking change. Existing local vector databases won't work with the new embeddings model. Those that want to keep the current database should not update to this commit.

Double check this project it have GPU..
https://github.com/lm-sys/FastChat

PR #254, opened 2023-05-17
Add .doc .ppt (Word and PowerPoint 97/2003 formats)

Fabio3rs:
To avoid manually converting documents saved in old format, add for the loader to load these as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ The supported extensions are:
 
    - `.csv`: CSV,
    - `.docx`: Word Document,
+   - `.doc`: Word Document,
    - `.enex`: EverNote,
    - `.eml`: Email,
    - `.epub`: EPub,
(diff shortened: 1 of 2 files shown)


mdeweerd:
I tried to keep the lists in alphabetical order - doc is before docx, and ppt before pptx.
It will avoid future false identifications of changes if someone decides to sort them.

Fabio3rs:
> Thanks for adding those - I was not sure that these filters work for doc/ppt as well.
> 
> I tried to keep the lists in alphabetical order - doc is before docx, and ppt before pptx. It will avoid future false identifications of changes if someone decides to sort them.

Thanks, I changed the order.

Seems to work fine here with the version in the requirements.txt, apparently the library converts the files into docx/pptx in background to use it.

imartinez (approved):

PR #255, opened 2023-05-17
Parallel document loading with multiprocessing Pool

Fabio3rs:
Faster loading of many documents by loading multiple documents in parallel using thread pool


ingest.py
```diff
@@ -65,7 +68,11 @@ def load_documents(source_dir: str) -> List[Document]:
         all_files.extend(
             glob.glob(os.path.join(source_dir, f"**/*{ext}"), recursive=True)
         )
-    return [load_single_document(file_path) for file_path in all_files]
+    pool = multiprocessing.Pool(os.environ.get('LOAD_DOCUMENTS_NUMBER_OF_THREADS', multiprocessing.cpu_count()))
```
watrgoat, line 71: I don't think it is recommended to use every thread possible, since it can cause a lockup, I would at least subtract 1 from the cpu_count() probably more depending on the number someone has on their system.

ingest.py
```diff
@@ -65,7 +68,11 @@ def load_documents(source_dir: str) -> List[Document]:
...
+    result = pool.map(load_single_document, all_files)
+    pool.close()
+    pool.join()
```
watrgoat, line 74: It is also typical to use `with` for anything like this for exception handling. example code would look like this:
```
with multiprocessing.Pool(os.environ.get('LOAD_DOCUMENTS_NUMBER_OF_THREADS', multiprocessing.cpu_count()-1)) as pool:
    pool.map(load_single_document, all_files)
```
besides that everything else looks good

watrgoat (changes requested):

ingest.py
Fabio3rs: Thanks, I did a commit fixing it

ingest.py
Fabio3rs: I kept the "- 1" from the example bellow.

I don't know if it's necessary to make a more complex code to calculate the number of threads, since the user can change the .env file, what do you think?

watrgoat: Yes I agree

watrgoat (approved):
Everything looks great!

Fabio3rs:
> Everything looks great!

Thanks

Was missing a int cast when the env file is used for the data. Can you re-review? Thanks.

watrgoat (approved):
Everything looks fixed and this should definitely improve the speed of file ingestion.

johnbrisbin:
You may find that any performance improvement you see will be very dependent on the type of media where the document files are located. 
That is, attempts to read multiple documents at the same time from rotating media (and to a lesser extent SSD) will tend to be much slower than reading them one at a time because with rotating media moving from one file (disk location) to another is very expensive. Similarly, most SSD devices, while much faster, will have similar issues because they have a limited number of buffers to handle concurrent IOs.
The part that can be profitably parallelized is the chunking which is a pure memory operation without a latency gotcha.

PulpCattel:
https://github.com/imartinez/privateGPT/pull/292 is merged in master so this PR is now obsolete. One thing we can take from this for the future is a user option (for both CLI and `.env`) to select the number of threads.

Fabio3rs:
> #292 is merged in master so this PR is now obsolete. One thing we can take from this for the future is a user option (for both CLI and `.env`) to select the number of threads.

Thanks. I think it's better to open a new PR with the user option feature.

PR #120, opened 2023-05-14
docker file and compose

JulienA:

PR #267, opened 2023-05-17
update Dockerfile

zxjason:
update Dockerfile


mdeweerd:
You may want to checkout #120 & https://github.com/mdeweerd/privateGPT/tree/cuda .

PR #284, opened 2023-05-18
Update ingest.py: Fix duplication of load_dotenv() and uncomment Docx2txtLoader

antsh3k:
This pull request addresses two specific changes:

1. Removal of `load_dotenv()` duplication: Previously, the `load_dotenv()` function was called twice in the code, leading to unnecessary duplication. In this pull request, I have removed the duplicated call and added it once at the beginning of the script to ensure proper loading of environment variables.

2. Uncommenting `# ".docx": (Docx2txtLoader, {})` line.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -5,6 +5,7 @@ from dotenv import load_dotenv
 
 from langchain.document_loaders import (
     CSVLoader,
+    Docx2txtLoader,
     EverNoteLoader,
     PDFMinerLoader,
     TextLoader,
--- a/ingest.py
+++ b/ingest.py
@@ -30,7 +30,7 @@ load_dotenv()
 # Map file extensions to document loaders and their arguments
 LOADER_MAPPING = {
     ".csv": (CSVLoader, {}),
-    # ".docx": (Docx2txtLoader, {}),
+    ".docx": (Docx2txtLoader, {}),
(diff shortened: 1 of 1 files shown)


ingest.py
```diff
@@ -30,7 +30,7 @@
 # Map file extensions to document loaders and their arguments
 LOADER_MAPPING = {
     ".csv": (CSVLoader, {}),
-    # ".docx": (Docx2txtLoader, {}),
+    ".docx": (Docx2txtLoader, {}),
```
PulpCattel, line 33: If you want to do this, can't judge, you have to add the `Docx2txtLoader` class to the `from langchain.document_loaders import (...)` import above, or the script won't run.

Also, I think you should either comment out the other `.docx` entry right below, or change it to another filetype.

PulpCattel (commented):
The `load_dotenv()` removal makes sense.

ingest.py
antsh3k: Ah you are right. UnstructuredWordDocumentLoader does make more sense.

Although The load_dotenv() removal makes sense. There is an existing PR [#268 ](https://github.com/imartinez/privateGPT/pull/268)

Thanks @PulpCattel

imartinez:
Merged https://github.com/imartinez/privateGPT/pull/268. Closing this one. Thanks!

PR #287, opened 2023-05-18
Add script to update the database for adding more documents

maozdemir:
Move environment variables to the global scope
Add a better check for vectorstore existence
Introduced a new function for better readability
Co-authored-by: Pulp <51127079+PulpCattel@users.noreply.github.com>

Recreating the PR https://github.com/imartinez/privateGPT/pull/201 due to the history getting way too messy (my frustration was beyond imagination...). Sorry for the inconvenience.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -24,6 +24,16 @@ from langchain.docstore.document import Document
 from constants import CHROMA_SETTINGS
 
 
+load_dotenv()
+
+
+# Load environment variables
+persist_directory = os.environ.get('PERSIST_DIRECTORY')
+source_directory = os.environ.get('SOURCE_DIRECTORY', 'source_documents')
+embeddings_model_name = os.environ.get('EMBEDDINGS_MODEL_NAME')
+chunk_size = 500
+chunk_overlap = 50
+    
 # Map file extensions to document loaders and their arguments
 LOADER_MAPPING = {
     ".csv": (CSVLoader, {}),
@@ -44,7 +54,6 @@ LOADER_MAPPING = {
 
 load_dotenv()
 
-
 def load_single_document(file_path: str) -> Document:
     ext = "." + file_path.rsplit(".", 1)[-1]
     if ext in LOADER_MAPPING:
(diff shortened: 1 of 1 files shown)


PulpCattel:
Tested and works. IMHO the `does_vectorstore_exist` function implementation is too fragile/hardcoded. I think even just checking if `persist_directory` is empty or not may be enough and much cleaner, but I'm not sure.

Thank you for working on this!

ingest.py
```diff
@@ -24,6 +24,16 @@
 from constants import CHROMA_SETTINGS
 
 
+load_dotenv()
```
imartinez, line 27: load_dotenv() is called in line 55 too. We should remove one of the two

ingest.py
```diff
@@ -55,37 +64,61 @@ def load_single_document(file_path: str) -> Document:
...
+    documents = load_documents(source_directory, ignored_files)
+    if not documents:
+        print("No new documents to load")
+        exit(0)
+    print(f"Loaded {len(documents)} new documents from {source_directory}")
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=chunk_size, chunk_overlap=chunk_overlap)
     texts = text_splitter.split_documents(documents)
-    print(f"Loaded {len(documents)} documents from {source_directory}")
-    print(f"Split into {len(texts)} chunks of text (max. {chunk_size} characters each)")
+    print(f"Split into {len(texts)} chunks of text (max. 500 tokens each)")
```
imartinez, line 91: instead of hardcoded 500, you could use the chunk_size variable

ingest.py
```diff
@@ -55,37 +64,61 @@ def load_single_document(file_path: str) -> Document:
...
+
+def does_vectorstore_exist(persist_directory: str) -> bool:
+    """
+    Checks if vectorstore exists
+    """
+    if os.path.exists(os.path.join(persist_directory, 'index')):
+         if os.path.exists(os.path.join(persist_directory, 'chroma-collections.parquet')) and os.path.exists(os.path.join(persist_directory, 'chroma-embeddings.parquet')):
+            list_index_files = glob.glob(os.path.join(persist_directory, 'index/*.bin'))
+            list_index_files += glob.glob(os.path.join(persist_directory, 'index/*.pkl'))
+            if len(list_index_files) == 4:
```
imartinez, line 102: This implementation is quite fragile, as @PulpCattel pointed out. It may break with a new version of Chroma.

imartinez (commented):
Just a couple comments. The solution looks great to me. Thanks a lot for this contribution!

ingest.py
maozdemir: What would be your suggeson for foolproofing the check for an existing index?

imartinez: @PulpCattel suggested checking if persist_directory is empty or not, but in my experience when ingest fails the directory contains 2 documents... so I can't think of a much better solution. Maybe using > 3 instead of == 4 makes it a little more flexible. But I don't love that solution either, so I'd be happy to go with this for the moment

PulpCattel: Yeah I think for now this can be a workable compromise.
If we keep this, a couple of small fixes:
* The second `if` from the top is over-indented.
* The two `path.join()` could become `path.join(persist_directory, 'index', '*.bin')`, i.e. split `'index/*.bin'` and `'index/*.pkl'`

imartinez:
@PulpCattel @maozdemir I updated the branch with main, applied the comments and merged it to main.

Issue #294, opened 2023-05-18
eml support


peterchanws:
I had issue ingest eml files.  #290
[pseudotensor](https://github.com/pseudotensor) kindly have written code for that https://github.com/h2oai/h2ogpt#readme
https://github.com/h2oai/h2ogpt/blob/main/gpt_langchain.py#L364-L375
. I don't know how to integrate the code into privateGPT. Could someone merge the code to privateGPT?
Thanks

mdeweerd:
Check if https://github.com/mdeweerd/privateGPT/pull/new/elm_extended works for you and comment on PR #295

PulpCattel:
@mdeweerd You meant #299?

mdeweerd:
Yes, sorry for the typo.

peterchanws:
Sorry, I am not a coder. Could someone help to review? Thanks.

peterchanws:
I used ChatGPT to review the code and I got "Overall, the code seems to implement a fallback mechanism to handle cases where the default loading method for UnstructuredEmailLoader fails to find HTML content in an email."

Issue #290, opened 2023-05-18
ValueError: text/html content not found in email


peterchanws:
**Describe the bug and how to reproduce it**
I put ~4000 eml files in source_document folder. Run ingest.py and got:

 File "/Users/pchan3/Desktop/privateGPT/ingest.py", line 78, in main
    documents = load_documents(source_directory)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pchan3/Desktop/privateGPT/ingest.py", line 65, in load_documents
    return [load_single_document(file_path) for file_path in all_files]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pchan3/Desktop/privateGPT/ingest.py", line 65, in <listcomp>
    return [load_single_document(file_path) for file_path in all_files]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pchan3/Desktop/privateGPT/ingest.py", line 53, in load_single_document
    return loader.load()[0]
           ^^^^^^^^^^^^^
  File "/Users/pchan3/miniconda3/lib/python3.11/site-packages/langchain/document_loaders/unstructured.py", line 70, in load
    elements = self._get_elements()
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/pchan3/miniconda3/lib/python3.11/site-packages/langchain/document_loaders/email.py", line 24, in _get_elements
    return partition_email(filename=self.file_path, **self.unstructured_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pchan3/miniconda3/lib/python3.11/site-packages/unstructured/partition/email.py", line 249, in partition_email
    raise ValueError(f"{content_source} content not found in email")
ValueError: text/html content not found in email 

**Environment (please complete the following information):**
 - OS / hardware: macOS 12.6.5 / Intel Xeon E5]
 - Python version [3.11.3]

pseudotensor:
We solved that already in h2oGPT.  See https://github.com/h2oai/h2ogpt

i.e.  https://github.com/h2oai/h2ogpt/blob/main/gpt_langchain.py#L364-L375

peterchanws:
How can I add it to privateGPT?

[...]

PR #299, opened 2023-05-18
Add fallback for plain elm #294 #290

mdeweerd:
To be validated by users.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -31,7 +31,7 @@ LOADER_MAPPING = {
     ".doc": (UnstructuredWordDocumentLoader, {}),
     ".docx": (UnstructuredWordDocumentLoader, {}),
     ".enex": (EverNoteLoader, {}),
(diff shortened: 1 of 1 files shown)


peterchanws (approved):

mdeweerd:
@imartinez I tested this with samples from https://github.com/mikel/mail/tree/master/spec/fixtures/emails/plain_emails .

The code also demonstrates how wrap an existing loader so that the existing LOADER_MAPPING dict can be maintained

imartinez (approved):
Looking really good. I think we should move custom loaders to their own files, but I think it is a good idea to keep this first one in ingest.py for the moment so people can see how to create custom loaders.

mdeweerd:
> I think we should move custom loaders to their own files

Certainly, but after adding #120 which moves the scripts to their own directory ;-), and maybe just one file such as "loaders.py".

PR #305, opened 2023-05-19
added CASALIOY ref

su77ungr:
**Add note inside README** 

references [CASALIOY (air-gapped toolkit)](https://github.com/su77ungr/CASALIOY) 

**Why _your_ community would benefit?**
We took some time and effort to supercharge this base idea of a private LLM client and solved dozens of your repo's **issues** also. Choosing different routes in things like vector storage, ingestion approach and GUI for good measure. Some but not all features:

- [x] Speed improvements through _qdrant_ and change in the base model
- [x] Blazing fast _parallelized ingestion_ and local HF tokenizer
- [x] _GUI_ built with streamlit to let your family chat in the browser (easy to embed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,5 +111,10 @@ When running a Mac with Intel hardware (not M1), you may run into _clang: error:
 
 If so set your archflags during pip install. eg: _ARCHFLAGS="-arch x86_64" pip3 install -r requirements.txt_
 
+## Other Alternatives (run locally)
+
+- [Casalioy](https://github.com/su77ungr/CASALIOY) toolkit with GUI (also fixed several issues of this repo)
+
+
 # Disclaimer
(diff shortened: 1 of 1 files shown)


imartinez:
@su77ungr thanks a lot for evolving the functionality and fixing issues! 

I won't reference your fork in the main readme because it would be unfair to the rest of contributors. There are 2K forks of this repo already, most of them adding different value on top of the base functionality of privateGPT. 

Please feel free to keep on contributing your improvements to the this main repo, or to keep on evolving your repo in a different direction. Your work is much appreciated either way! This is the beauty of Open Source in my opinion, you can choose the path forward.

As a not very relevant side-note, some of the improvements you mention are already present in the main branch of this repo.

su77ungr:
> 

This has nothing to do with fairness. We are not at a competition, are we? 

Also if you prefer keeping 70+ people hanging that's an interesting choice to make in favor of "fairness" 

Also most of these 2k "forks" are not maintained and do not contribute anything. There's not a single one of them with 100+ stars. 

That's all. Closed for me.

PR #312, opened 2023-05-19
Added script arguments to disable printing of LLM stdout Stream and source documents

abhiruka:
Sometimes may not need the source or the LLM stdout stream to be shown, just need the answer.

Refactored main function to take hide_source and mute_stream parameters for controlling output. 
Added argparse for command-line argument parsing. 
StreamingStdOutCallbackHandler and source document display are now optional based on user input.  

kept the current behaviour as the default when not parsing any arguments. 

Also, updated README.md to reflect these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,25 +83,7 @@ Type `exit` to finish the script.
 
 
 ### Script Arguments
-The script also supports optional command-line arguments to modify its behavior:
-
-- `--hide-source` or `-S`: Use this flag to disable printing of the source documents used for answers. By default, the source documents are printed.
-  
-```shell
-python privateGPT.py --hide-source
-```
-
-- `--mute-stream` or `-M`: Use this flag to disable LLM standard output streaming response, which by default prints progress to the console.
-
-```shell
-python privateGPT.py --mute-stream
-```
-
-You can combine these options if needed:
-
-```shell
-python privateGPT.py --hide-source --mute-callback
-```
+The script also supports optional command-line arguments to modify its behavior. You can see a full list of these arguments by running the command ```python privateGPT.py --help``` in your terminal
 
 
 # How does it work?
--- a/README.md
+++ b/README.md
@@ -81,6 +81,29 @@ Note: you could turn off your internet connection, and the script inference woul
 
 Type `exit` to finish the script.
 
+
+### Script Arguments
+The script also supports optional command-line arguments to modify its behavior:
+
+- `--hide-source` or `-S`: Use this flag to disable printing of the source documents used for answers. By default, the source documents are printed.
+  
+```shell
+python privateGPT.py --hide-source
+```
+
+- `--mute-stream` or `-M`: Use this flag to disable LLM standard output streaming response, which by default prints progress to the console.
+
+```shell
+python privateGPT.py --mute-stream
+```
+
+You can combine these options if needed:
+
+```shell
+python privateGPT.py --hide-source --mute-callback
+```
+
+
 # How does it work?
 Selecting the right local models and the power of `LangChain` you can run the entire pipeline locally, without any data leaving your environment, and with reasonable performance.
 
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -5,6 +5,7 @@ from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
 from langchain.vectorstores import Chroma
 from langchain.llms import GPT4All, LlamaCpp
 import os
+import argparse
 
 load_dotenv()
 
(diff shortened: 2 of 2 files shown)


privateGPT.py
```diff
@@ -17,12 +18,13 @@
 
 from constants import CHROMA_SETTINGS
 
-def main():
+def main(hide_source=False, mute_stream=False):
```
PulpCattel, line 21: As per the other comment, this should remain just `def main():`
Also note that `hide_source` and `mute_stream` are already set to `False` by default because of `action='store_true'`

privateGPT.py
```diff
@@ -31,27 +33,42 @@ def main():
...
         print("\n> Answer:")
         print(answer)
-        
-        # Print the relevant sources used for the answer
-        for document in docs:
-            print("\n> " + document.metadata["source"] + ":")
-            print(document.page_content)
+
+        # Print the relevant sources used for the answer, if source is True
+        if not hide_source and docs:
```
PulpCattel, line 54:
```suggestion
        if docs:
```
Nit:
This should be enough, we already check `hide_source` above, if `docs` is `None` we know there are no documents.

PulpCattel (commented):
Thank you, the CLI interface is very nice to have, at least having `python3 myscript.py -h` working is great. Here or in another PR we should add the same to `ingest.py`.

Also, I think we should at least add the same options we have in the `.env` file, tho this can be done in a follow-up PR too. And viceversa, these new options should maybe go into the `.env` config file too.

PulpCattel:
BTW, @imartinez thank you very much for working on this project.

privateGPT.py
```diff
@@ -31,27 +33,42 @@ def main():
...
+                        help='Use this flag to disable the streaming StdOut callback for LLMs.')
+
+    return parser.parse_args()
+
 
 if __name__ == "__main__":
-    main()
+    # Parse the command line arguments
+    args = parse_arguments()
+    main(hide_source=args.hide_source, mute_stream=args.mute_stream)
```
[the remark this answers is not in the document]
abhiruka: Thanks for spotting that , yes indeed it will easier to maintain in the long run and it the correct way to do it. I have updated accordingly.

privateGPT.py
```diff
@@ -31,27 +33,42 @@ def main():
...
-            print(document.page_content)
+
+        # Print the relevant sources used for the answer, if source is True
+        if not hide_source and docs:
+            for document in docs:
+                print("\n> " + document.metadata["source"] + ":")
+                print(document.page_content)
+
+def parse_arguments():
+    parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
abhiruka: @PulpCattel i have added a description @imartinez i used the existing description in the readme. 

> _privateGPT: Ask questions to your documents without an internet connection,
> using the power of LLMs._

privateGPT.py
abhiruka: Yep, moved it inside the main as per the previous comment and remove this, keeping ```def main():``` as is

README.md
```diff
@@ -81,6 +81,29 @@ Note: you could turn off your internet connection, and the script inference woul
...
+- `--mute-stream` or `-M`: Use this flag to disable LLM standard output streaming response, which by default prints progress to the console.
+
+```shell
+python privateGPT.py --mute-stream
+```
+
+You can combine these options if needed:
+
+```shell
+python privateGPT.py --hide-source --mute-callback
```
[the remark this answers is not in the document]
abhiruka: Yep true, made it more concise as suggested.  

> The script also supports optional command-line arguments to modify its behavior. You can see a full list of these arguments by running the command ```python privateGPT.py --help``` in your terminal

privateGPT.py
abhiruka: yep agreed, Opted to remove the ```if``` and assign an empty list instead of ```None``` seems more cleaner and simpler.

abhiruka:
@PulpCattel thanks for the review much appreciated. @imartinez thanks, happy to pitch in. I have updated accordingly.

README.md
```diff
@@ -81,6 +81,11 @@ Note: you could turn off your internet connection, and the script inference woul
 
 Type `exit` to finish the script.
 
+
+### Script Arguments
```
PulpCattel, line 85: Nit:
Script Arguments sounds a bit weird, some options like `CLI` or `Command line interface` or something maybe sounds a bit better.

```suggestion
### CLI

File: README.md [hunk 7]
```diff
@@ -81,6 +81,11 @@ Note: you could turn off your internet connection, and the script inference woul
...
+The script also supports optional command-line arguments to modify its behavior. You can see a full list of these arguments by running the command ```python privateGPT.py --help``` in your terminal
```
PulpCattel, line 86:
```suggestion
The script also supports optional command-line arguments to modify its behavior. You can see a full list of these arguments by running the command ```python privateGPT.py --help``` in your terminal.
```
Nit: missing punctuation mark.

PulpCattel (commented):
Looks good. Just left some nits, feel free to ignore.

One thing we could start doing more consistently is squashing PRs commits, to keep the git history clean. @imartinez do we care about this?

imartinez (approved):
This is a really good contribution, thanks both! I may expose in the Readme this option more explicitly in a future commit.

abhiruka:
@imartinez thanks mate and keep up the good work.

Issue #257, opened 2023-05-17
Progress Bar for injest


apcameron:
Having a Progress Bar or the percentage displayed would be helpful on injest.py

scorpion44:
I am up to~500 CPU hours on my xeon server (32 core) ingesting 100mb of text, no idea what's happening, or if something is actually being done....

isthisausername2:
> I am up to~500 CPU hours on my xeon server (32 core) ingesting 100mb of text, no idea what's happening, or if something is actually being done....

There is a new update that significantly reduces the time to ingest.

scorpion44:
> > I am up to~500 CPU hours on my xeon server (32 core) ingesting 100mb of text, no idea what's happening, or if something is actually being done....
> 
> There is a new update that significantly reduces the time to ingest.

Was the update in the last 25 hours, as that's when I started???

isthisausername2:
I believe it was introduced when this PR was merged. (Cant remember if it was this exact one but it was around this time)
If you are on the version with this update i would recommend leaving it for a few more hours (just to see if it actually does finish) and if it dosent i would redownload the repo and model.  

https://github.com/imartinez/privateGPT/commit/355b4be7c0972f71208251a14f47d739f8456fb5

gardner:
Something like this works in `main()`

```
if len(texts) > 100:
    batch_size = len(texts) / 100
    batches = [texts[i:i + batch_size] for i in range(0, len(texts), batch_size)]
    for batch in tqdm(batches, desc="Processing batches"):
        db = Chroma.from_documents(
            batch, ef, persist_directory=persist_directory)
else:
    db = Chroma.from_documents(
        texts, ef, persist_directory=persist_directory)
```

gardner:
I also noticed ingest.py would load all of the documents into memory before starting to add them to the index and create emdeddings. I modified it to load documents and add them to the index one-by-one. This reduces the mempory overhead when importing large amounts of data. My use case only loads txt. This is what my `ingest.py` looks like now:


```
#!/usr/bin/env python3
import nltk
import os
import glob
from typing import List
from dotenv import load_dotenv
from multiprocessing import Pool
from tqdm import tqdm
import chromadb
from chromadb.utils import embedding_functions

from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain.docstore.document import Document
from langchain.document_loaders import TextLoader


load_dotenv()

#  Load environment variables
[...]

PR #292, opened 2023-05-18
Optimize load_documents function with multiprocessing

jiangzhuo:
This pull request introduces multiprocessing in the `load_documents()` function to improve performance by utilizing multiple cores for document loading.

**Key Changes**

- The `load_documents()` function now uses Python's multiprocessing module, allowing for concurrent document loading across multiple CPU cores.
- This modification takes advantage of machines with multiple cores and/or hyper-threading capabilities, leading to significant potential improvements in document loading times.
- It enhances performance, especially when dealing with a large number of documents.

This update is expected to significantly speed up the document loading process, contributing to overall system efficiency and user experience. Please review and provide your feedback.

***^^^This commit message is generated by ChatGPT^^^***

PR #315, opened 2023-05-19
Update ingest.py

virtualdvid:
implemented tqdm progress bar.

This is a simple way to solve this issue: https://github.com/imartinez/privateGPT/issues/257

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -22,6 +22,7 @@ from langchain.vectorstores import Chroma
 from langchain.embeddings import HuggingFaceEmbeddings
 from langchain.docstore.document import Document
 from constants import CHROMA_SETTINGS
+from tqdm import tqdm
 
 
 class MyElmLoader(UnstructuredEmailLoader):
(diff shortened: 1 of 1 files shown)


josmas:
Hey @virtualdvid have you tried this with more than one file? I am not seeing what I would expect to see.

I have clipped the top part, but it basically shows 0% for all the first 385 documents I have, then for the last one it jumps to 100%. But then it stays much longer in the last steps doing the DB part, even though the progress bar is already at 100.

```
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
  0%|                                                                              | 0/11 [00:00<?, ?it/s]
100%|██████████████████████████████████████████████████████████████████| 386/386 [00:03<00:00, 106.41it/s]
Loaded 386 documents from source_documents
Split into 2950 chunks of text (max. 400 characters each)
Using embedded DuckDB with persistence: data will be stored in: db

```

As an added note, would you consider adding tqdm to requirements? this only works because some other package seems to depend on it, so it is already installed.

PulpCattel (commented):
The progress bar is nice to have, tho the real deal would be having a progress bar for the ingestion itself (which is I think what https://github.com/imartinez/privateGPT/issues/257 is about), but IDK if that's even possible with current libraries.

Also I agree with this:
> As an added note, would you consider adding tqdm to requirements? this only works because some other package seems to depend on it, so it is already installed.

imartinez:
Thanks for the contribution! As @PulpCattel pointed out, this is already covered by https://github.com/imartinez/privateGPT/pull/292/files. Closing this PR

PR #318, opened 2023-05-20
Fix future incompatibility from langchain Chroma

johnbrisbin:
Cleanup #5003 which removes the from_documents functionality which conflicted with another method of the same name.
Replaced from_documents with from_texts. See:
https://github.com/hwchase17/langchain/pull/5003

It looks like this will go through and they have marked the change as breaking existing code when it goes into the main branch and is released.

In more detail, the from_documents method is removed and another incompatible prototype is all that remains (in vectorstores.py)
Apparently, the from_documents implementation used in privateGPT was a convenience to connect the output of text_splitters (Documents) which contain both the text and metadata to Chroma from_texts which expects two lists, one with strings and one with dicts (metadata). The code used in the old implementation of from_documents to split the Document list into two list is inserted into ingest.py (main) to allow use of the from_texts method.
This approach is compatible with both the old and prospective versions of langchains support for the Chroma db.

This is my first pull request, so feel free to correct my technique.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -79,7 +79,9 @@ def main():
     chunk_overlap = 50
     documents = load_documents(source_directory)
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=chunk_size, chunk_overlap=chunk_overlap)
-    texts = text_splitter.split_documents(documents)
+    text_docs = text_splitter.split_documents(documents)
+    texts = [doc.page_content for doc in text_docs]
+    metadatas = [doc.metadata for doc in text_docs]
     print(f"Loaded {len(documents)} documents from {source_directory}")
     print(f"Split into {len(texts)} chunks of text (max. {chunk_size} characters each)")
 
@@ -87,7 +89,7 @@ def main():
     embeddings = HuggingFaceEmbeddings(model_name=embeddings_model_name)
     
     # Create and store locally vectorstore
-    db = Chroma.from_documents(texts, embeddings, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)
+    db = Chroma.from_texts(texts=texts, embedding=embeddings, metadatas=metadatas, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)
     db.persist()
     db = None
 
--- a/ingest.py
+++ b/ingest.py
@@ -79,9 +79,9 @@ def main():
     chunk_overlap = 50
     documents = load_documents(source_directory)
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=chunk_size, chunk_overlap=chunk_overlap)
-    text_docs = text_splitter.split_documents(documents)
-    texts = [doc.page_content for doc in text_docs]
-    metadatas = [doc.metadata for doc in text_docs]
+    chunks = text_splitter.split_documents(documents)
+    texts = [doc.page_content for doc in chunks]
+    metadatas = [doc.metadata for doc in chunks]
     print(f"Loaded {len(documents)} documents from {source_directory}")
     print(f"Split into {len(texts)} chunks of text (max. {chunk_size} characters each)")
 
--- a/ingest.py
+++ b/ingest.py
@@ -89,7 +89,7 @@ def main():
     embeddings = HuggingFaceEmbeddings(model_name=embeddings_model_name)
     
     # Create and store locally vectorstore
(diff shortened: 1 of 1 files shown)


ingest.py
```diff
@@ -79,15 +79,17 @@ def main():
     chunk_overlap = 50
     documents = load_documents(source_directory)
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=chunk_size, chunk_overlap=chunk_overlap)
-    texts = text_splitter.split_documents(documents)
+    text_docs = text_splitter.split_documents(documents)
+    texts = [doc.page_content for doc in text_docs]
+    metadatas = [doc.metadata for doc in text_docs]
```
PulpCattel, line 84: It seems a bit wasteful to loop two times through `text_docs` (NIT: BTW we could just call it `docs`), instead of just once. On the other hand, this code comes from `langchain` directly, so maybe it's worth keeping it as is just for convenience sake.

ingest.py
```diff
@@ -79,15 +79,17 @@ def main():
...
     print(f"Loaded {len(documents)} documents from {source_directory}")
     print(f"Split into {len(texts)} chunks of text (max. {chunk_size} characters each)")
 
     # Create embeddings
     embeddings = HuggingFaceEmbeddings(model_name=embeddings_model_name)
     
     # Create and store locally vectorstore
-    db = Chroma.from_documents(texts, embeddings, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)
+    db = Chroma.from_texts(texts=texts, embedding=embeddings, metadatas=metadatas, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)
```
PulpCattel, line 92:
```suggestion
    db = Chroma.from_texts(texts, embeddings, metadatas=metadatas, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS)
```
This works, so maybe no need to change it.

PulpCattel (commented):
I can't comment on what `Chroma` and `langchain` are doing, hopefully this breaking change won't hit us anyway for a long time. Do they plan to replace this convenience method with another convenience method?

In any event, this change can't hurt, the code change is simple and the commit message is very clear. I checked the new code added is exactly the same from the `from_documents()` `classmethod`.

Briefly tested and it works no problem.

ingest.py
johnbrisbin: Among the python priesthood, list comprehensions are considered more concise and readable than 'for' loops. I don't see any claim that they are more efficient, though. Were it not for the 'original' mechanism, I would have written a 'for' loop.

On the var name, I am not feeling much joy with calling the chunks produced by the splitter 'docs' or, originally, 'documents' because, despite having the type they are very distinct from the original documents. Text_docs was a compromise of function 'text' and type 'doc'. On further consideration, I would choose chunks for clarity.

While I am whining about things, I don't think we are getting very good service from the loader-splitter combinations in terms of metadata. The splitters take the raw text and apply the same metadata from the document to each chunk. That is fine for a two or ten page document, but it is not very helpful to know that the relevant passage is somewhere in a 500 page book when it comes time to refer to, quote that snippet or create a citation.
The loaders know about pages, but don't add anything to the output to identify where the boundaries are and the splitters don't even add the offset into the text conversion to the metadata.
And yes, I have quite a few 500 page books in my digital collection.
</whining>

PulpCattel: Oh I prefer list comprehension too, but intuitively it doesn't make sense. Tho, you make a very good point about performance. Because of how Python works, it's quite possibly much faster than even just a single `for` loop .

`for chunk in chunks` works very well for me too.

johnbrisbin: @PulpCattel Thanks for your attentions. I realized that time has marched on and I probably need to merge these changes into the now much different ingest.py on the main branch.
That will give me the chance to introduce new errors in the process but I will try to keep it to a minimum. I don't know if I need to put a pause on this request beyond this notice or do anything else while I work. Not a big deal but a pretty big change for the appearance of the pull request.

johnbrisbin: I explored the more current code and it is not really compatible with this strategy. However, it offers much greater opportunity for dealing with this case and related issues.

johnbrisbin:
In exploring the current ingest code I have concluded that this approach is not appropriate, but another, better approach is possible based on the current ingest alog.
I am abandoning and closing this request in favor of the new, better one to come.

PR #324, opened 2023-05-20
refactor: Update document loading and vectorstore creation process

maozdemir:
- Rename the function `does_vectorstore_exist` to `ensure_integrity` for its new purpose.
- We have to ensure that the database is intact, in order to use it.
- Refactor main function to simplify the code. Apparently the Chroma database will be updated if it exists, or will be created if it does not exist in a given path.


ingest.py
```diff
@@ -126,37 +126,33 @@ def process_documents(ignored_files: List[str] = []) -> List[Document]:
...
             list_index_files = glob.glob(os.path.join(persist_directory, 'index/*.bin'))
             list_index_files += glob.glob(os.path.join(persist_directory, 'index/*.pkl'))
-            # At least 3 documents are needed in a working vectorstore
+            # More than 3 documents are needed in a working vectorstore
             if len(list_index_files) > 3:
-                return True
-    return False
+                return True # Vectorstore exists and is valid
+        return False # Vectorstore exists but is not valid
+    return True # Vectorstore does not exist yet
```
PulpCattel, line 141: This doesn't seem necessary, but you could add the printing inside this function.

PulpCattel (commented):
`ensure_integrity` is duplicated which is not great, if we want to do the check in both scripts we should add a `util.py` module or something. Same thing for the `if not ensure_integrity(persist_directory):` check. All the printing could be done inside `ensure_integrity` directly.

The simplification might be okay, tho I don't mind as it was before, it seems more robust and allows better logging.

maozdemir:
> `ensure_integrity` is duplicated which is not great, if we want to do the check in both scripts we should add a `util.py` module or something. Same thing for the `if not ensure_integrity(persist_directory):` check. All the printing could be done inside `ensure_integrity` directly.
> 
> The simplification might be okay, tho I don't mind as it was before, it seemed more robust and allowed better logging.

These should satisfy your concerns.

PulpCattel:
I'm not entirely convinced about this change, not sure if it's necessary and I can see benefits with as is on master.

Let's wait to see what plans @imartinez has for something like this. Adding some extra utility modules was mentioned in some other PRs, and there are various ways to do it.

.gitignore
```diff
@@ -7,6 +7,7 @@ models/
 # Local Chroma db
 .chroma/
 db/
+source_documents/
```
PulpCattel, line 10: But we have a `source_documents` by default no? Is the idea to remove the default source too?

maozdemir: You do have a point, I was switching branches and that just slipped lol, should change it to directory contents instead of the whole folder.

PulpCattel: I think if we want to have a default source file, like the state of the union or anything else, it should not be in `.gitignore` either.

PR #327, opened 2023-05-20
Use os.walk, fix some issues identified with pylint

mdeweerd:
This replaces #109.  os.walk is faster than globbing.

I also ran pylint on the code and fixed some other issues.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -124,16 +124,14 @@ def load_documents(
     with Pool(processes=os.cpu_count()) as pool:
         results = []
         with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_progress_bar else nullcontext()) as pbar:
+            progress_write=tqdm.write if use_progress_bar else print
             for file_path, doc in pool.imap_unordered(load_single_document, filtered_files):
                 results.append(doc)
                 if use_progress_bar:
                     pbar.update()
                 if report_processed_files:
                     msg = f"Processed '{file_path}'"
-                    if use_progress_bar:
-                        tqdm.write(msg)
-                    else:
-                        print(msg)
+                    progress_write(msg)
 
     return results
 
--- a/ingest.py
+++ b/ingest.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 import os
-import glob
 import re
 import sys
 from typing import List, Optional
@@ -146,13 +145,15 @@ def does_vectorstore_exist(persist_path: str) -> bool:
     """
     Checks if vectorstore exists
     """
-    if os.path.exists(os.path.join(persist_path, 'index')):
-        if os.path.exists(os.path.join(persist_path, 'chroma-collections.parquet')) and os.path.exists(os.path.join(persist_path, 'chroma-embeddings.parquet')):
-            list_index_files = glob.glob(os.path.join(persist_path, 'index/*.bin'))
-            list_index_files += glob.glob(os.path.join(persist_path, 'index/*.pkl'))
-            # At least 3 documents are needed in a working vectorstore
+    index_path = os.path.join(persist_path, 'index')
+    if os.path.exists(index_path):
+        chroma_collections_path = os.path.join(persist_path, 'chroma-collections.parquet')
+        chroma_embeddings_path = os.path.join(persist_path, 'chroma-embeddings.parquet')
+        if os.path.exists(chroma_collections_path) and os.path.exists(chroma_embeddings_path):
+            list_index_files = [f for f in os.listdir(index_path) if f.endswith(('.bin', '.pkl'))]
             if len(list_index_files) > 3:
                 return True
+
     return False
 
 def main():
--- a/ingest.py
+++ b/ingest.py
@@ -2,7 +2,7 @@
 import os
 import sys
 from contextlib import nullcontext
-from typing import List, Optional
+from typing import List, Optional, Tuple
 from dotenv import load_dotenv
 from multiprocessing import Pool
 from tqdm import tqdm
@@ -82,12 +82,12 @@ LOADER_MAPPING = {
 }
 
 
-def load_single_document(file_path: str) -> Document:
+def load_single_document(file_path: str) -> Tuple[str, Document]:
     ext = "." + file_path.rsplit(".", 1)[-1]
 
     loader_class, loader_args = LOADER_MAPPING[ext]
     loader = loader_class(file_path, **loader_args)
-    return loader.load()[0]
+    return (file_path, loader.load()[0],)
 
 
 def load_documents(
@@ -125,12 +125,11 @@ def load_documents(
     with Pool(processes=os.cpu_count()) as pool:
         results = []
         with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_progress_bar else nullcontext()) as pbar:
-            for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
+            for file_path, doc in pool.imap_unordered(load_single_document, filtered_files):
                 results.append(doc)
                 if use_progress_bar:
                     pbar.update()
                 if report_processed_files:
-                    file_path = filtered_files[i]
                     msg = f"Processed '{file_path}'"
                     if use_progress_bar:
                         tqdm.write(msg)
--- a/ingest.py
+++ b/ingest.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import sys
+from contextlib import nullcontext
 from typing import List, Optional
 from dotenv import load_dotenv
 from multiprocessing import Pool
@@ -95,7 +96,7 @@ def load_documents(
     report_processed_files: bool = True,
     report_skipped_files: bool = True,
     report_ignored_files: bool = True,
-    use_process_bar: bool = True,
+    use_process_bar: bool = False,
 ) -> List[Document]:
 
     """
@@ -123,7 +124,7 @@ def load_documents(
 
     with Pool(processes=os.cpu_count()) as pool:
         results = []
-        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_process_bar else None) as pbar:
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_process_bar else nullcontext()) as pbar:
             for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
                 results.append(doc)
                 if use_process_bar:
--- a/ingest.py
+++ b/ingest.py
@@ -96,7 +96,7 @@ def load_documents(
     report_processed_files: bool = False,
     report_skipped_files: bool = True,
     report_previously_ingested_files: bool = False,
-    use_process_bar: bool = True,
+    use_progress_bar: bool = True,
 ) -> List[Document]:
 
     """
@@ -124,10 +124,10 @@ def load_documents(
 
     with Pool(processes=os.cpu_count()) as pool:
         results = []
-        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_process_bar else nullcontext()) as pbar:
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_progress_bar else nullcontext()) as pbar:
             for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
                 results.append(doc)
-                if use_process_bar:
+                if use_progress_bar:
                     pbar.update()
                 if report_processed_files:
                     file_path = filtered_files[i]
--- a/ingest.py
+++ b/ingest.py
@@ -116,7 +116,7 @@ def load_documents(
                     filtered_files.append(file_path)
                 else:
                     if report_previously_ingested_files:
-                        print(f"Ignored '{file_path}' (ignore list)")
+                        print(f"Ignored '{file_path}' (already ingested, ignored if updated!)")
             else:
                 if report_skipped_files:
                     file_path = os.path.join(root, file)
--- a/ingest.py
+++ b/ingest.py
@@ -92,18 +92,18 @@ def load_single_document(file_path: str) -> Document:
 
 def load_documents(
     source_dir: str,
-    ignored_files: Optional[List[str]] = None,
-    report_processed_files: bool = True,
+    ingested_files: Optional[List[str]] = None,
+    report_processed_files: bool = False,
     report_skipped_files: bool = True,
-    report_ignored_files: bool = True,
-    use_process_bar: bool = False,
+    report_previously_ingested_files: bool = False,
+    use_process_bar: bool = True,
 ) -> List[Document]:
 
     """
     Loads all documents from the source documents directory, ignoring specified files
     """
-    if ignored_files is None:
-        ignored_files : List[str] = []
+    if ingested_files is None:
+        ingested_files : List[str] = []
 
     filtered_files = []
     extensions = tuple(LOADER_MAPPING.keys())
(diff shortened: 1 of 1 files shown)


ingest.py
```diff
@@ -91,49 +93,63 @@ def load_single_document(file_path: str) -> Document:
...
-        exit(0)
+        sys.exit(0)
     print(f"Loaded {len(documents)} new documents from {source_directory}")
     text_splitter = RecursiveCharacterTextSplitter(chunk_size=chunk_size, chunk_overlap=chunk_overlap)
     texts = text_splitter.split_documents(documents)
     print(f"Split into {len(texts)} chunks of text (max. {chunk_size} tokens each)")
     return texts
 
-def does_vectorstore_exist(persist_directory: str) -> bool:
+def does_vectorstore_exist(persist_path: str) -> bool:
```
PulpCattel, line 145: I suppose you changed the name because it shadows the global one. But note that since it's a global, we don't need to pass it to this function at all. So you could just remove the parameter and leave the name unchanged.

That being said, since using globals is not great and eventually I think we'll change that, this change still makes sense.

ingest.py
```diff
@@ -91,49 +93,63 @@ def load_single_document(file_path: str) -> Document:
...
     Checks if vectorstore exists
     """
-    if os.path.exists(os.path.join(persist_directory, 'index')):
-        if os.path.exists(os.path.join(persist_directory, 'chroma-collections.parquet')) and os.path.exists(os.path.join(persist_directory, 'chroma-embeddings.parquet')):
-            list_index_files = glob.glob(os.path.join(persist_directory, 'index/*.bin'))
-            list_index_files += glob.glob(os.path.join(persist_directory, 'index/*.pkl'))
+    if os.path.exists(os.path.join(persist_path, 'index')):
+        if os.path.exists(os.path.join(persist_path, 'chroma-collections.parquet')) and os.path.exists(os.path.join(persist_path, 'chroma-embeddings.parquet')):
+            list_index_files = glob.glob(os.path.join(persist_path, 'index/*.bin'))
+            list_index_files += glob.glob(os.path.join(persist_path, 'index/*.pkl'))
```
PulpCattel, line 152: I guess we could use `walk` here too and remove `glob` entirely, or `listdir`.

PulpCattel (commented):
The `pylint` changes I quite like.

I personally prefer the implementation from https://github.com/imartinez/privateGPT/pull/109/. I don't see a reason to add regex stuff when it's such a simple check.

ingest.py
mdeweerd: It's not good practice to use globals - I try to use globals only when they are real constants.   There's generally a reason down the road to not use globals.

mdeweerd:
Regarding the regex: it's a way to allow the user to specify a regex on the CLI somewhere down the road.  It's probably also faster than using "in <list>" on each file as that list will grow longer and longer.

PulpCattel:
Oh, I just noticed (it happens on master too) that we are silently ignoring every unknown/missing file type.

```python3
def load_single_document(file_path: str) -> Document:
    ext = "." + file_path.rsplit(".", 1)[-1]
    if ext in LOADER_MAPPING:
        loader_class, loader_args = LOADER_MAPPING[ext]
        loader = loader_class(file_path, **loader_args)
        return loader.load()[0]
    raise ValueError(f"Unsupported file extension '{ext}'")
```

`load_single_document()` looks like this, but we have already filtered all the `file_path` in `load_documents()`.
So this `raise` is de facto unreachable. and the `if` is redundant.

Since `load_single_document` is parallelized, it could make sense to do the file type check there.
Alternatively, we could keep the filtering and remove the `if` and the `raise` from `load_single_document()` and instead maybe add a `print` during the `walk` to warn the user that we skipped a file.

mdeweerd:
> So this `raise` is de facto unreachable. and the `if` is redundant.

IMHO a method function is written with regards to its caller, but with regards to its function/contract.
If this function is reused for another reason, this is important.

Suppose that we allow the user to specify a list of documents explicitly without walking the tree.  Then trying to load an unsupported document raises an exception.

ingest.py
mdeweerd: @PulpCattel Optimized in the update.

PulpCattel:
> @PulpCattel IMHO a method function is written with regards to its caller, but with regards to its function/contract.
If this function is reused for another reason, this is important.

Sure, I can see this argument, tho I still don't like that at the moment it's completely useless, while at the same time being so easy to change when needed.

In any case, even if we want to keep it as future-proof thing, we should not silently ignore files. The user won't expect that.

mdeweerd:
> given how easy it's to change if/when needed.

In my experience that's more overall effort and often forgotten/not done until a bug appears that takes longer to identify than when a warning/exception was present.

In a way, the first Ariane 5 was blown up because an overflow exception was not caught which was not necessary for Ariane 4.  But they reused the code from Ariane 4 in Ariane 5 and decided not to test it because it was already field-proven through Ariane 4.

> In any case, even if we want to keep it as future-proof thing, we should not silently ignore files. The user won't expect that.

The files are not truely silently ignored as the supported files are listed in the README.  The other options would be to list all ignored files (and console I/O is relatively costly and polluting), and to do this only when a CLI option is active, or to count the number of ignored files as well and indicate "XX/YY files will be ingested".

PulpCattel:
> In a way, the first Ariane 5 was blown up because an overflow exception was not caught which was not necessary for Ariane 4. But they reused the code from Ariane 4 in Ariane 5 and decided not to test it because it was already field-proven through Ariane 4.

I mean, you are talking about a security critical, massive project, while here we are dealing with some Python scripts (not even a library yet) that already are not focused on security, dead simple, and with nothing at stake.
Also note that even after removing the `raise` and `if` the result (of a new function coded without checking what the function it calls does) would be (virtually always) the same, the code would raise an exception (mostly `KeyError`?) if an unknown/missing file type is found, the only difference would be a worse error message.

Nonetheless, I appreciate this level of enthusiasm, and I'm okay leaving it there if you feel strongly about it.

> The files are not truely silently ignored as the supported files are listed in the README.

They are silently ignored. The user might not even realize a typo in the file extension, or that some files amongst many others miss it.

> The other options would be to list all ignored files (and console I/O is relatively costly and polluting), and to do this only when a CLI option is active, or to count the number of ignored files as well and indicate "XX/YY files will be ingested".

Yes, these are the options I was thinking about. Telling the user which files are being ignored doesn't seem polluting, it seems very helpful. Also, it should be a rare thing to begin with, and if it really fills the user screen then that's a good thing, or we could have skipped many of his files without him even realizing (if he put the files in the `source` folder, supposedly he expects them to be ingested). 

Just keeping a count is a much simpler option, but it doesn't help if the user has to then go and search amongst possibly many, many files the ones that are missing the extension.

I think we should warn the user by default, and possibly have a CLI option to disable the check if the user doesn't care.

mdeweerd:
1. Programming principles
    Big or small, principles are the same.  Adding an exception is a no-brainer for me, not adding it is just counterintuitive and knowing it's more work in the end.
    Checking "two times" is also good for "Defense in Depth".  There is no impact on performance here because normally that line is never reached and there is no extra check.
2. Verbosity.
    I prefer the lowest verbosity by default.  Providing a count can already trigger the user to question which files were skipped and increase verbosity if he wants to know.

PulpCattel:
> Checking "two times" is also good for "Defense in Depth". There is no impact on performance here because normally that line is never reached and there is no extra check.

The "two times" (the `if`) is the performance impact, and is present every single time.
I.e., the "security/defense in depth" benefit of this, in this context, for this function, doesn't seem a good justification to add de-facto dead code and lower performance. It's a trade-off that just doesn't convince me.
Again, not a big deal either way for me.

The `raise` thing is much easier to go along with, as you say it's never reached and therefore it has no impact, so it almost becomes a stylistic choice (note that by adding it in advance now, you may end up doing *more* work, because you cannot fully predict the refactors, necessities, and requirements you'll have in the future).

> I prefer the lowest verbosity by default. Providing a count can already trigger the user to question which files were skipped and increase verbosity if he wants to know.

Implementing both the count and the full "log", with the log off by default, would not be my preference but it seems a good compromise.
The argument in favor of more verbosity by default would be that ingesting files is the whole point of this script, so it seems a big deal if we are not ingesting files. I.e., we want to be **sure** to trigger the user to question, especially if many files have been skipped.

mdeweerd:
New commit, 

- Use endswith instead of regex (preferred #109 by @imartinez),
- Remove extension test in load_single_document (@PulpCattel [comment](https://github.com/imartinez/privateGPT/pull/327#issuecomment-1556215277),
- Inform/Warn the user by default (@PulpCattel [comment](https://github.com/imartinez/privateGPT/pull/327#issuecomment-1556189523)."


I'll leave it up to somebody else to add the argument parsing and option setting for disabling the reports.

One can either create a new PR to [imartinez:main](https://github.com/imartinez/privateGPT) or to [mdeweerd:os_walk](https://github.com/mdeweerd/privateGPT/tree/os_walk) which I'll merge to update this PR #327 .

In case a new PR to [imartinez:main](https://github.com/imartinez/privateGPT) is created, I'll close this one.

ingest.py
```diff
@@ -83,60 +83,89 @@ def load(self) -> List[Document]:
...
-        with tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) as pbar:
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_process_bar else None) as pbar:
             for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
                 results.append(doc)
-                pbar.update()
+                if use_process_bar:
+                    pbar.update()
+                if report_processed_files:
+                    file_path = filtered_files[i]
+                    print(f"Processed '{file_path}'")
```
PulpCattel, line 133: I think the printing inside the `tqdm` is creating some problems.
I see this:
```sh
Skipping 'source_documents/state_of_the_union.tx' - unmatched extension
Loading new documents:   0%|                              | 0/4 [00:00<?, ?it/s]Processed 'source_documents/bar.csv'
Processed 'source_documents/foo.txt'
Processed 'source_documents/A/barfoo.csv'
Processed 'source_documents/B/foobar.csv'
Loading new documents: 100%|█████████████████████| 4/4 [00:00<00:00, 584.88it/s]
```

ingest.py
```diff
@@ -83,60 +83,89 @@ def load(self) -> List[Document]:
...
+                        print(f"Ignored '{file_path}' (ignore list)")
+            else:
+                if report_skipped_files:
+                    file_path = os.path.join(root, file)
+                    print(f"Skipping '{file_path}' - unmatched extension")
 
     with Pool(processes=os.cpu_count()) as pool:
         results = []
-        with tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) as pbar:
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_process_bar else None) as pbar:
```
PulpCattel, line 126: With `use_process_bar=False` this crashes. Maybe just let the progress bar always on? It doesn't seem worth the hassle to make it optional.
BTW, is `process` intentional or did you mean `progress`?

PulpCattel (commented):
I like it. Tested a bit, left some suggestions below.
The only thing that seems excessive to me is to print all the processed files, that seems expected behavior and no need to tell users, especially by default.

ingest.py
```diff
@@ -83,60 +83,89 @@ def load(self) -> List[Document]:
...
+    if ignored_files is None:
+        ignored_files : List[str] = []
+
+    filtered_files = []
+    extensions = tuple(LOADER_MAPPING.keys())
+
+    for root, _, files in os.walk(source_dir):
+        for file in files:
+            if file.endswith(extensions):
+                file_path = os.path.join(root, file)
```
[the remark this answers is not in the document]
mdeweerd: The full file_path is not always needed in the else expression, so to avoid a useless joing for those that are not reporting the unmatched extensions, the filepath is computed only when needed..

ingest.py
```diff
@@ -83,60 +83,89 @@ def load(self) -> List[Document]:
...
+                if file_path not in ignored_files:
+                    filtered_files.append(file_path)
+                else:
+                    if report_ignored_files:
+                        print(f"Ignored '{file_path}' (ignore list)")
```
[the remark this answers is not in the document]
mdeweerd: New push, I renamed "ignored" to "ingested" and changed the default reporting levels.

ingest.py
mdeweerd: This is an expected "misbehavior".

Which is why these are options, there should CLI argument parsing that disables the progress bar when file reporting is active, unless the user forces the progress bar.

As indicated above, the CLI argument parsing is left to another volunteer.

Still, with the progress bar the user can still see how much progress there is.

ingest.py
mdeweerd: "process bar" and "progress bar" both seem to be used in the wild, but here it should be "progress" which seems more popular -> renamed.

The crash is also fixed.

ingest.py
PulpCattel: I think using `tqdm.write(f"Processed '{file_path}'")` instead of `print()` should fix it.

ingest.py
```diff
@@ -83,60 +84,89 @@ def load(self) -> List[Document]:
...
+    extensions = tuple(LOADER_MAPPING.keys())
+
+    for root, _, files in os.walk(source_dir):
+        for file in files:
+            if file.endswith(extensions):
+                file_path = os.path.join(root, file)
+                if file_path not in ingested_files:
+                    filtered_files.append(file_path)
+                else:
+                    if report_ingested_files:
```
PulpCattel, line 118: Typo here, `report_ingested_files` is undefined, you called it `report_previously_ingested_files`.

BTW, sorry if it wasn't clear, I meant clarify to the user. E.g., instead of `(ignore list)` something like `(already ingested)`?

mdeweerd: 1. I forgot to rename both locations.
2. I updated the message to the user.

ingest.py
mdeweerd: Pushed the change.

ingest.py
```diff
@@ -83,60 +84,93 @@ def load(self) -> List[Document]:
...
-        with tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) as pbar:
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_progress_bar else nullcontext()) as pbar:
             for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
                 results.append(doc)
-                pbar.update()
+                if use_progress_bar:
+                    pbar.update()
+                if report_processed_files:
+                    file_path = filtered_files[i]
+                    msg = f"Processed '{file_path}'"
```
PulpCattel, line 134: Due to `pool.imap_unordered()` above, I think the `file_path` processed is not necessarily the one at index `i`.
Probably this is intentional just to print all of them conveniently, just pointing it out, and in case we could change to `imap()`.

PulpCattel (commented):
Nice! Thank you, done a little more testing and works well.

ingest.py
mdeweerd: I made a test case, the order is indeed not guaranteed.

I updated the code so that the paths can be reported as they are treated.

ingest.py
```diff
@@ -81,62 +82,93 @@ def load(self) -> List[Document]:
...
-            for i, doc in enumerate(pool.imap_unordered(load_single_document, filtered_files)):
+        with (tqdm(total=len(filtered_files), desc='Loading new documents', ncols=80) if use_progress_bar else nullcontext()) as pbar:
+            for file_path, doc in pool.imap_unordered(load_single_document, filtered_files):
                 results.append(doc)
-                pbar.update()
+                if use_progress_bar:
+                    pbar.update()
+                if report_processed_files:
+                    msg = f"Processed '{file_path}'"
+                    if use_progress_bar:
```
tardigrde, line 133: could be a 1-liner  `tqdm.write(msg) if use_progress_bar else print(msg)`?

tardigrde: True. It's a tradeoff. Your choice.

mdeweerd:
I am closing this PR as I am deleting my fork.

PR #91, opened 2023-05-13
ADD UI Chat simple

vnk8071:
Thanks @imartinez for great repo about private data and LLM offline.
- Make the conservation of chat
- Can edit prompt for instruction
- Can edit config model, number context
- Clear history chat

PR #329, opened 2023-05-20
Added updated UI using Gradio

roy-mootsana:


sime2408:
Already requested in #91. Maybe it's better to do a code review there.

vnk8071:
Thanks @sime2408 for mentioning my PR#91. This UI of @roy-mootsana can collaborate with me to make a better version based on the user's choice.

imartinez:
@vnk8071 @roy-mootsana @sime2408 thanks a lot for this, and sorry for not prioritizing this into main. Please don't think I'm ghosting this idea, on the contrary, I'm trying to build a roadmap that plays well with the rest of initiatives in the space.

Logical steps:
- we'll need to make PrivateGPT a library that decouples the user interface from the ingestion and inference logic
- we'll need to make sure we support our own UIs within this project while also being open to other's such as Quivr

I'd like to hear your thoughts, maybe in the Discussions tab, or maybe in the Discord we are making available really really soon ;)

vnk8071:
Thanks, @imartinez for the quick reply. The open source you mention is here [https://github.com/StanGirard/quivr](https://github.com/StanGirard/quivr) right? We can make it together if you would like to share your UI. And maybe I think we should find a way for user convenience such as a quick demo without ingestion.

PR #365, opened 2023-05-22
Added bat and sh files for install, ingest and start

ilian6806:
I think that this is good practice and a more convenient way to install and start tools like this, especially for non-programmers.

### DIFF
--- a/ingest.sh
+++ b/ingest.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+echo "Starting ingest..."
+source ./venv/Scripts/activate
+python ingest.py
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+echo "Creating venv..."
+mkdir venv
+python -m venv venv
+echo "Activating venv..."
+source ./venv/Scripts/activate
+echo "Installing requirements..."
+pip install -r requirements.txt
+echo "Successfully installed!"
--- a/ingest.bat
+++ b/ingest.bat
@@ -1,0 +1,5 @@
+@echo off
+echo Starting ingest...
+call .\venv\Scripts\activate.bat
+python ingest.py
+pause
--- a/install.bat
+++ b/install.bat
@@ -1,0 +1,10 @@
+@echo off
+echo Creating venv...
+mkdir venv
+python -m venv venv
+echo Activating venv...
+call .\venv\Scripts\activate.bat
+echo Installing requirements...
+pip install -r requirements.txt
+echo Successfully installed!
+pause
--- a/start.bat
+++ b/start.bat
@@ -1,0 +1,4 @@
+@echo off
+call .\venv\Scripts\activate.bat
+python privateGPT.py
+pause
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+source ./venv/Scripts/activate
+python privateGPT.py


ingest.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
PulpCattel, line 1:
```suggestion
#!/usr/bin/env bash
```
I think this is more flexible, you can do the same in the other file too.

ingest.sh
```diff
@@ -0,0 +1,4 @@
...
+echo "Starting ingest..."
+source ./venv/Scripts/activate
```
PulpCattel, line 3:
```suggestion
source ./venv/bin/activate
```
Why `Scripts`? Same thing in the other places.

ingest.sh
```diff
@@ -0,0 +1,4 @@
...
+python ingest.py
```
PulpCattel, line 4:
```suggestion
python3 ingest.py
```
This would be preferred everywhere I think.

install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+echo "Creating venv..."
```
PulpCattel, line 2: If we want to do this, we should add `venv` to the requirements. In some systems (Ubuntu/Debian for example) it's not installed by default.

install.sh
```diff
@@ -0,0 +1,9 @@
...
+mkdir venv
```
PulpCattel, line 3: This is not needed I think. `venv` creates the folder for us.

install.sh
```diff
@@ -0,0 +1,9 @@
...
+python -m venv venv
+echo "Activating venv..."
+source ./venv/Scripts/activate
+echo "Installing requirements..."
+pip install -r requirements.txt
```
PulpCattel, line 8:
```suggestion
pip3 install -r requirements.txt
```
Recommended.

PulpCattel (commented):
Only looked at the `.sh` file, can't comment on the Windows stuff.
For now the installation is so simple that maybe these extra files are not needed. On the other hand, doing the virtualenv for users is very helpful I think.

If we add these files, we should also mention them in the README
BTW, you should make all of them executable by default (`chmod +x`).

ingest.sh
ilian6806: Because I did this on a windows pc haha. Sorry, fixed it.

ingest.sh
ilian6806: My be it is a good idea to pass venv and python paths as optional params as in Stable Diffusion Web UI for example.

install.sh
ilian6806: I think this is the best practice, but this is just my suggestion.

maozdemir:
Honestly, this feels like creating a script for just one-liner. The setup process is not that complex to require a setup and start script.
Also, a problem with this sort of implementation is the outcome of people with multiple Python versions installed.

@PulpCattel how would a user benefit from this implementation? I'm struggling to see the actual point.

ilian6806:
Thanks for the feedback @PulpCattel ! I addressed all of these. 

Personally, I think virtual env is a must for things like this. The problem with different python versions and venv dirs can be solved easy with arguments.

I just wanted to share some practices from my experience with stable diffusion development, because I think, there are a lot of non-programmer users now and every manual install is kind of a barrier to them.

install.sh
```diff
@@ -0,0 +1,9 @@
...
+echo "Successfully installed!"
```
maozdemir, line 12: This might get better if there was a prompt to ask if the user wants to download a model, and suggest some of the common models and download with curl or equivalent (apparently curl is available on Windows 10 too, see https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409)

install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
maozdemir, line 1:
```suggestion
#!/bin/bash
mv example.env .env
```

maozdemir (changes requested):

PulpCattel:
> how would a user benefit from this implementation?

By using a virtualenv. I'm pretty sure most people using this repo are installing all the requirements on their base Python installation. We already have problems of people messing up dependencies because of that.

In my experience too, using `venv` is pretty standard and strictly better (except UX-wise, it's too hard for most users and on their own they will never do it).
Now, if it's necessary to add a script instead of just some more info in the README, I'm not sure either (as I said in my first post). but surely can't be terrible. Also, as you noted, the install script could do a few more things for the user automatically.
Two things about that:
* Wouldn't it be better to `cp` instead of `mv` the example file? So that the user still retains the original copy. We could change this also in the README. Not a big deal.
* The extra things, like downloading models, etc, could be done in follow-ups I think. I'd also prefer, in case they become more complicated, to have Python scripts for that, rather than bash.

ingest.sh
PulpCattel: This is not done in the other two scripts (`start` and `install`).

ingest.sh
PulpCattel: This also needs to change in the other scripts too.

install.sh
```diff
@@ -0,0 +1,13 @@
...
+mv example.env .env
+
+echo "Creating venv..."
+python3 -m venv venv
+
+echo "Activating venv..."
+source ./venv/Scripts/activate
```
PulpCattel, line 8: One thing here is that if, for some reason (should be rare), this fails (say `python3 -m venv` is not available on the machine), then we go ahead and install the requirements anyway, we could maybe error or prompt the user in that case.

ingest.sh
ilian6806: I changed it

install.sh
ilian6806: If we do not want to require users to use venv, I think prompt for global installation is ok

install.sh
johnbrisbin: I tend to agree about asking. Sometimes people are on expensive connections and should have the chance to put off large downloads until it costs less or nothing.

johnbrisbin:
I tend to like the idea of some kind on installer not because it is difficult right now (pretty close though) but because it will only get more complex. Keeping installation sensible enough that a script or simple installer can do it is a good way of keeping it on the radar. A successful project must run pretty much out of the box because everyone who can't get it to work goes away mad or sad.
I have seen several videos in recent days about privateGPT and they were generally quite favorable. From those, you are going to get more and less capable people trying to install it and, if they fail, those reviews are going to turn sour fast.

PR #383, opened 2023-05-22
fix exit bug

bearney74:
when someone types exit, the program does not exit.  This pull request fixes that bug, and allows any input that starts with exit (or quit) in upper/lower, or mixed case to terminate the program.


johnbrisbin:
Modal text interfaces are a pain, but I would take the traditional ^C over starting to pick commands out of innocent text. Soon enough there will be proper UI to offer quit options in a way that won't cause unintended failures. We should have no magic words in the text.
If you must do something like this, have exactly one word and the query must be that word exactly with no other characters. Have two words and you will soon have ten. Still a bad idea.

kenancum:
I had the same problem. I am using the app on WSL. My Ubuntu os froze after I type exit. I fixed the problem by terminating the process after the while loop.

PR #387, opened 2023-05-22
typo: Change `pip` to `pip3` in README.md

maozdemir:
On some environments with both Python 2.x and 3.x installed, `pip` will be used for 2.x.


imartinez (approved):
Thanks!

PR #408, opened 2023-05-23
feat: add support to use OpenAI for llm use

mudler:
This adds the possibility to use OpenAI as a llm to answer questions. This makes it possible to use privateGPT locally as well with projects like https://github.com/go-skynet/LocalAI


janvarev:
Looks like my PR too :) https://github.com/imartinez/privateGPT/pull/325 
I have added KoboldAPI endpoint too.

PR #421, opened 2023-05-23
Update ingest.py

amrith-am:


johnbrisbin:
How do you justify this change?
It will certainly make the ingested database much smaller and reduce the RAM footprint, but the size of the chunks is also reflected in the similarity nearest neighbors used to feed the LLM through the prompt. Making them that large will mean that you will not be able to pass more than one near neighbor to the LLM for the prompt to operate on and generate a result.  Because only a single item is passed to the LLM, the quality of the result will suffer and may be out in left field more often than not.
Our models have pretty small working contexts (context windows). The Open AI contexts are huge by comparison and permit much larger chunking than we can support reasonably.
I suggested increasing the chunk size to 1000 a couple of days ago and I am not certain that wasn't too much.

amrith-am:
What was the window size of the gola model

PR #423, opened 2023-05-23
Improve Handling of Unsupported File Extensions

markmcwong:
Improves the handling of files with unsupported extensions during the document ingestion process. 

Previously, if a file with an unsupported extension was encountered, the `load_single_document` function would raise a `ValueError`, potentially interrupting the ingestion process.

With this change, the `load_single_document` function will simply print a warning message and return `None` when it encounters a file with an unsupported extension. The `load_documents` function has also been updated to filter out these `None` results. This means that the ingestion process will continue with the rest of the files even if some files have unsupported extensions.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -88,7 +88,8 @@ def load_single_document(file_path: str) -> Document:
         loader = loader_class(file_path, **loader_args)
         return loader.load()[0]
 
-    raise ValueError(f"Unsupported file extension '{ext}'")
+    print(f"Warning: Unsupported file extension '{ext}' for file {file_path}. Skipping this file.")
+    return None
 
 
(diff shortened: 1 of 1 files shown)


markmcwong:
Thanks for pointing it out! In this case since `load_documents` already pre filtered the known extensions, perhaps in this case double checking the document loader and the format of `Loader_Mapping` is more appropriate:

- Added validation to check the loader mapping format and raise an error if it does not match the expected `(document loader, arguments)` structure.
- Improved validation to ensure that the loader class is callable and has the `load` method.

PulpCattel:
`LOADER_MAPPING` is hardcoded as a global dictionary. I think what you are adding would be more useful as a unit test. Inside a function doesn't seem great.

PR #460, opened 2023-05-24
feat: Get answers using preferred number of chunks

maozdemir:
The current implementation uses 4 chunks by default. This commit adds the functionality of changing the number of chunks to be used in the generation of answers.

Answers #416


imartinez (approved):
Nice contribution!

maozdemir:
@imartinez it's still `TARGET_SOURCE_DOCUMENTS` instead of `TARGET_SOURCE_CHUNKS` in example.env.

imartinez:
Could you open a PR to fix it? I'm away from keyboard @maozdemir

PR #475, opened 2023-05-25
Simple Interactive Ingest - Allows use of multiple sets of sources

id-b3:
Added a simple command line interactive prompt for the ingest script to enable the user to select a set of documents.
This addresses the issue of having to delete the source folder and db if they want to restart with another project.

See the updated README.md for the new approach in the ingest section.

The solution still uses the dotenv approach.


.gitignore
```diff
@@ -7,6 +7,7 @@ models/
 # Local Chroma db
 .chroma/
 db/
+dbs/
```
sime2408, line 10: Maybe better to add `.gitignore` & `.gitkeep` inside the `dbs` folder and remove this from project `.gitignore`? An example of .gitignore:
```txt
*
!.gitignore
!.gitkeep
```

ingest.py
```diff
@@ -139,6 +139,86 @@ def does_vectorstore_exist(persist_directory: str) -> bool:
...
+    selected_directory_path (str): The path of the selected directory.
+    selected_db_path (str): The path of the database directory for the selected directory.
+    """
+
+    def _display_directories():
+        """
+        This function displays the list of existing directories in the ./sources directory.
+        """
+        print("Existing directories in ./sources:")
+        directories = os.listdir("./sources")
```
sime2408, line 159: This will list those annoying folders generated by Mac for example: `.DS_Store`, or if you have `.gitkeep` or similar files inside it will also list them

ingest.py
```diff
@@ -164,4 +244,7 @@ def main():
 
 
 if __name__ == "__main__":
+
+    source_directory, persist_directory = prompt_user()
```
sime2408, line 248: `db.persist()` doesn't want to store parquet files when I run it for the 1st time, only if a folder is previously created in the `dbs`, then it creates embeddings as it should. 
I suspect that the rest of the program is not aware that a new directory is created ?? Hm... This is working ok for you?

ingest.py
```diff
@@ -139,6 +139,86 @@ def does_vectorstore_exist(persist_directory: str) -> bool:
...
+                    print(f"Selected directory: {selected_directory_path}")
+                    break
+            except (ValueError, IndexError):
+                print("Invalid directory number. Please try again.")
+        elif choice == "2":
+            new_directory_name = input("Enter the name for the new directory: ")
+            selected_directory_path, selected_db_path = _create_directory(new_directory_name)
+            input("Place your source material into the new folder and press enter to continue...")
+            break
+        elif choice == "q":
```
sime2408, line 214: Suggestions to add choice option num 3 :
```python

print(f"3. Current source_directory is {source_directory}. Do you want to use it?")

..

elif choice == "3":
  return source_directory, persist_directory

```

sime2408: Even though I would extract this input prompting to separate method and add timeout that option 3 is the default one executed if there's no input:

```python
    def _get_user_choice(timeout):
        print("Select an option or 'q' to quit:")
        print("1. Select an existing directory")
        print("2. Create a new directory")
        print(f"3. Use current source_directory: {source_directory}")
        inputs = [sys.stdin]
        readable, _, _ = select.select(inputs, [], [], timeout)
        if readable:
            return sys.stdin.readline().strip()
        return "3"
```
code around:
```
import sys
import select
...
def prompt_user(timeout=10):
...
choice = _get_user_choice(timeout)
```

README.md
```diff
@@ -70,9 +80,9 @@ Using embedded DuckDB with persistence: data will be stored in: db
 Ingestion complete! You can now run privateGPT.py to query your documents
 ```
 
-It will create a `db` folder containing the local vectorstore. Will take 20-30 seconds per document, depending on the size of the document.
-You can ingest as many documents as you want, and all will be accumulated in the local embeddings database.
-If you want to start from an empty database, delete the `db` folder.
+It will create a subfolder in the `dbs` folder containing the local vectorstore. Will take 20-30 seconds per document, depending on the size of the document.
+You can ingest as many documents as you want, and all will be accumulated in the selected embeddings database.
+If you want to start from an empty database, delete the subfolder inside the `dbs` folder, or create a new one using the inject.py script.
```
sime2408, line 85: `inject.py -> ingest.py`

ingest.py
sime2408: Found out where's the problem. It is in `constants.py` file, where: 

```python
CHROMA_SETTINGS = Settings(
    chroma_db_impl='duckdb+parquet',
    persist_directory=PERSIST_DIRECTORY,
    anonymized_telemetry=False
)
```
Is always holding the previous `PERSIST_DIRECTORY` from the env variable. And, if you change the directory of ingest it will write chroma files to the old folder from `.env` variable. So the idea is to remove this constant from here and make it dynamic on change because `set_key('.env', 'YOUR_ENV_VAR', path)` will only store / change value in .env file when ingestion is done

.gitignore
id-b3: I would agree, but at the moment I'm trying to follow the current layout of the project which added the original `db` folder to the gitignore, so I'm mirroring this.
I think a project structure refactor would probably be best done in a separate set of commits to simplify things.

ingest.py
id-b3: Thanks for catching that possibility, I've also added logic to only list directories.
Tested against the source directory containing files, dotfiles and dotfolders.

README.md
id-b3: Thanks! Corrected now!

ingest.py
id-b3: Good idea, I agree with this approach, added and tested, seems to work well.

ingest.py
id-b3: It's an interesting bug, it doesn't seem to arise for me starting with a clean clone of the repo, but I can see that it could arise if things change without calling ingest.py before inference.

I'm not familiar with updating class parameters on environmental changes, I tried a few methods and can't seem to get the CHROMA_SETTINGS parameter to update. Maybe you have an approach?

Here is what I tried doing in the `constants.py` file:

```python
import os
from dotenv import load_dotenv
from chromadb.config import Settings

class DynamicSettings(Settings):
    @property
    def persist_directory(self):
        return self.get_persist_directory()

    def get_persist_directory():
        load_dotenv(override=True)
        return os.environ.get('PERSIST_DIRECTORY')

# Define the Chroma settings
CHROMA_SETTINGS = DynamicSettings(
    chroma_db_impl='duckdb+parquet',
    anonymized_telemetry=False,
)
```

ingest.py
```diff
@@ -139,6 +141,99 @@ def does_vectorstore_exist(persist_directory: str) -> bool:
...
+        print("1. Select an existing directory")
+        print("2. Create a new directory")
+        print(f"3. Use current source_directory: {source_directory}")
+        inputs = [sys.stdin]
+        readable, _, _ = select.select(inputs, [], [], timeout)
+        if readable:
+            return sys.stdin.readline().strip()
+        return "3"
+
+    def _display_directories():
```
sime2408, line 169: Nice way of displaying sub-directories in 4 columns if you have a lot of them, also I colored print so it's more visible

```python
import textwrap
import math
```

```python
def _display_directories():
        """
        This function displays the list of existing directories in the ./source_documents directory.
        """
        print(f"\n\033[94mExisting directories in ./source_documents:\033[0m")
        directories = sorted((f for f in os.listdir("./source_documents") if not f.startswith(".")), key=str.lower)

        # Set the desired column width and the number of columns
        column_width = 30
        num_columns = 4

        # Calculate the number of rows needed based on the number of directories
        num_rows = math.ceil(len(directories) / num_columns)

        # Print directories in multiple columns
        for row in range(num_rows):
            for column in range(num_columns):
                # Calculate the index of the directory based on the current row and column
                index = row + column * num_rows

                if index < len(directories):
                    directory = directories[index]
                    wrapped_directory = textwrap.shorten(directory, width=column_width - 1, placeholder="...")
                    print(f"{index+1:2d}. {wrapped_directory:{column_width}}", end="")

            print()  # Print a new line after each row

        return directories
```

id-b3: The colouring is a nice touch, but I think for now the columns are a bit too much. When testing this in a small terminal, or resizing it, the columns aren't nicely spaced leading to issues with readability. I think handling all those cases might increase the overall complexity of the interface, at which point it might be worthwhile investing time in a nice GUI instead, what do you think?

sime2408: GUI is definitely a better idea, but then on GUI, you should consider manipulating the change of the dbs sub-directory you want to ask questions to. Also, what if you want to combine them to have joint answers from multiple databases? Maybe something like this but I am not sure. This would be really nice feature to combine topics (folder of books) from the UI:

```python
db = Chroma(persist_directory=chroma_directory, embedding_function=embedding)

db.add_documents(documents=texts1)
db.add_documents(documents=texts2)
```

ingest.py
id-b3: Thanks for the tips, I ended up changing CHROMA_SETTINGS from a constant to a function which would dynamically update the persist directory when called, and CHROMA_SETTINGS occurences in the code. From my testing this seems to solve the error that you've been getting.

sime2408: Yea, I was confused why is this happening to me until I changed it at both places, you know that other place is here:

```python
Chroma(persist_directory=persist_directory........
```

so maybe if we commit to setting persistence inside CHROMA_SETTINGS, might help ... but glad it work now

PR #490, opened 2023-05-26
Refactor file extension retrieval using os.path.splitext()

cobanov:


ingest.py
```diff
@@ -82,7 +82,7 @@ def load(self) -> List[Document]:
 
 
 def load_single_document(file_path: str) -> Document:
-    ext = "." + file_path.rsplit(".", 1)[-1]
+    ext = os.path.splitext(file_path)[-1]
```
sime2408, line 85: Can you explain why is this a better approach?
By the way, you could lowercase it in case of uppercase extensions like .PDF etc:

`ext = (os.path.splitext(file_path)[-1]).lower()`

sime2408 (approved):

ingest.py
cobanov: Both approaches are valid, but the second one is better. The first approach uses the rsplit() method to split the file path at the last occurrence of the dot character. This is a bit inefficient because it has to scan the entire file path to find the dot character. The second approach uses the splitext() method, which is specifically designed to split a file path into its root and extension. This is more efficient, because it only has to scan the file path until it finds the first dot character.

In addition, the second approach is more readable, because it is more explicit about what it is doing. The first approach is a bit more cryptic, because it is not immediately clear what the rsplit() method is doing.

Yes, you are right about lowercase. Since it was not present in the previous version of the code, I did not include it. With this correction, it will be much safer.

ice0 (approved):

PR #526, opened 2023-05-29
feat: make it a library and a cli

### LINKED ISSUE #525: make it a library with a CLI
**Is your feature request related to a problem? Please describe.**
Simplify UX and use it as a library in other projects (e.g. clients) 

**Describe the solution you'd like**
A simple refactor as CLI + Python Library.

**Describe alternatives you've considered**
No other alternatives were currently explored.

**Additional context**
Original scripts `ingest.py` and `privateGPT.py` should keep the same interface to not break current usage.

emilianobonassi:
Fixes #525 

Keep the same interface for `ingest.py` and `privateGPT.py` but refactored to use a library and a CLI.

I've built this, to speed up my own usage and to prepare for building a specific client. Sharing with the community thinking it will be useful for you too.

With this PR, `privateGPT.py` becomes a Click-based CLI exposing the following commands:

- `chat`, the default command when the CLI is invoked without command that runs the original privateGPT scripts
- `ingest`, executes the logic for ingesting documents
- `init`, this is new and it's useful to bootstrap a new env, copying the default configuration and downloading the default model
- `reset`, just drop the database (instead of dropping manually the folder)`

Example of launching `privateGPT.py --help`

```
Usage: privateGPT.py [OPTIONS] COMMAND [ARGS]...

  Ask questions to your documents without an internet connection, using the
  power of LLMs.

Options:
  --help  Show this message and exit.

Commands:
  chat    Query your documents.
  ingest  Ingest your documents.
  init    Inititalize with default settings.
  reset   Reset your database.
```

NB: it's the beginning of making a library, a few things like `setup.py` and others are missing but could be added after this.


maozdemir (approved):
LGTM

privateGPT/ingestor.py
```diff
@@ -0,0 +1,169 @@
...
+def load_single_document(file_path: str) -> Document:
+    ext = "." + file_path.rsplit(".", 1)[-1]
+    if ext in LOADER_MAPPING:
+        loader_class, loader_args = LOADER_MAPPING[ext]
+        loader = loader_class(file_path, **loader_args)
+        return loader.load()[0]
+
+    raise ValueError(f"Unsupported file extension '{ext}'")
+
+def load_documents(source_dir: str, ignored_files: List[str] = []) -> List[Document]:
```
Kondrahin, line 80: To avoid further bugs, it is possible better to write `ignored_files: List[str] = None` (also in `_process_documents`) and then make a check
```python
if ignored_files is None:
        ignored_files = []
```
See important warning in [Python tutorial](http://docs.python.org/tutorial/controlflow.html#default-argument-values).

RonquilloAeon:
I like your architecture.

PR #558, opened 2023-05-30
if you accept this, .xlsx able

Midmost:
Hi it's the first time I'm trying to give pull request to someone
so please don't be mad whether I'm right or wrong....

Thanks for good source, good luck!

Sincerely, Momo

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -6,6 +6,9 @@ from dotenv import load_dotenv
 from multiprocessing import Pool
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


maozdemir (changes requested):
I doubt this is necessary since the Unstructured-IO library supports natively, despite not being included in the langchain environment. You could try implementing that instead.

Midmost:
OK thanks 😊

2023년 6월 2일 (금) 오전 9:52, maozdemir ***@***.***>님이 작성:

PR #576, opened 2023-05-31
Add GUI interface

REFRIED-BoolEANS:
Just a simple GUI made with tkinter.  The only problem is that I've never used github before so I actually gave the app a new name and even made an icon using dall-e (because I didn't know a pull request was a thing).  I can change the name, the readme, and use a different icon if you want to use the code, which does still need some work.

(or if you want to make this a library like someone else said, I can make this my own project if you are ok with that)

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -81,15 +81,14 @@ LOADER_MAPPING = {
 }
 
 
-def load_single_document(file_path: str) -> Document:
+def load_single_document(file_path: str) -> List[Document]:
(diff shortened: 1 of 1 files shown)


REFRIED-BoolEANS:
> If you create a new fork and add back the code with just the files for the GUI the PR has a chance of being approved. The way it is you're replaced a lot of references to the original instructions and naming in files like README.

Yeah, like I said I can change it all if needed.  I'll work on that, then do I create a new pull request or add it in here somehow?

PR #586, opened 2023-06-02
update with GUI code

REFRIED-BoolEANS:
This adds a chat interface with tabs to add files, create new files, and adds the potential for a much more user-friendly interface.  It definitely could use some improvement, but is a decent starting place in my opinion.  (I have only changed the code, added an icon, and haven't even messed with the readme)


danbiocchi:
Screenshot?

REFRIED-BoolEANS:
Nothing amazing, but it works.  For some reason on the 'Create new file' tab you have to make the window slightly larger to see the place to name the file you have to make the window larger.  I think that can be fixed pretty easily though.  Clicking the add-file button opens your file explorer so you can choose a file.
image
image
image

Issue #627, opened 2023-06-05
Incremental training (indigesting) of the data


chenle02:
**Is your feature request related to a problem? Please describe.**
This is not a problem but an additional feathure. 

**Describe the solution you'd like**
One may constantly adding more training documents into the source folder. When one run indigest.py, would it be possible to speed it up to add only the newly added documents? If would be nice, it can also detect if some old documents has been removed and update the database accordingly.

Thanks a lot~!

armoliss:
That was the wrong commit, I will update it soon!

pseudotensor:
Yes, this is good feature.  You can add to privateGPT how h2oGPT does it, which also allows update from CLI or running case from UI.

https://github.com/h2oai/h2ogpt/blob/main/gpt_langchain.py#L122-L154

Note it's important to not just add new docs, but delete the prior instances.  That block shows how to do that.

From h2oGPT, in UI, one just adds a new file into the originally-specified folder and click on "refresh sources" to update the db.  I use hash to detect file changes and add new files.  If files are deleted, however, I don't remove anything from db.  I think that's  most common desired behavior:

![image](https://github.com/imartinez/privateGPT/assets/2249614/2637d2c3-4e3d-4048-8306-9d42d720a718)

PR #629, opened 2023-06-05
feat: Implement incremental document ingestion

armoliss:
- Added a document tracking mechanism to keep track of previously ingested documents as referenced in #627 
- Modify `load_documents()` to filter out already ingested documents
- Handle removal of documents and update the database accordingly
- Improve progress tracking and logging during incremental ingestion

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -5,6 +5,7 @@ from typing import List
 from dotenv import load_dotenv
 from multiprocessing import Pool
 from tqdm import tqdm
+from msg_parser.msg_parser import MsOxMessage
 
 from langchain.document_loaders import (
     CSVLoader,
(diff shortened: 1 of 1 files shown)


abalib:
@armoliss  Incremental ingestion is useful.  Why did you close it?

abalib:
Thanks. I fetched your commit and noticed the bug too.   I think the "removal" feature should be optional.  May be you can add a flag from command line to enable removal?   Here is why by example.  If I dump 3000 pdfs in to ./source_documents, ingest.py doesn't complete; probably runs out of memory.   My workaround is to call ingest.py multiple times with a 30-50 pdfs per call, **removing** the processed pdf afterwards and on to the next batch.  If you start removing the missing documents from DB, then it's going to break this batching method

PR #521, opened 2023-05-29
VERY BIG performance improvement and beautiful features

DanielusG:
- Fixed an issue that made the evaluation of the user input prompt extremely slow, this brought a monstrous increase in performance, about 5-6 times faster.
- Added a script to install CUDA-accelerated requirements
- Added the OpenAI model (it may go outside the scope of this repository, so I can remove it if necessary)
- Added some additional flags in the .env
- Changed the embedder template to a better performing template 
- Bumped some versione, like llama-cpp-python and langchain (WARNING with the new version of llama.cpp, models are more powerful, but old models are now completely incompatible, if necessary, you can downgrade)
- I removed the state-of-union example file because someone who does not see it might leave it there and it would bother them during queries
- Added auto translation (Perhaps it should be removed because it uses an Internet connection)

PR #649, opened 2023-06-06
Fix slow gen

DanielusG:
Taken only the performance improvement modification from #521
I have explained in detail what each parameter added to the .env file does, but I feel I have been too verbose, any suggestions would be appreciated!


example.env
emrul, line 1: Small observation - the example file still has `MODEL_TYPE` set to `GPT4All` and these settings only apply to `LlamaCpp`. Maybe an example.gpu.env file would be more suitable? Rest of the patch looks really clean and I like the level of documentation for each setting.

sime2408: Or just:

```python
if model_type == "gpt4all":
    if gpu_is_enabled:
        logging.warn("GPU is enabled, but GPT4All does not support GPU acceleration. Please use LlamaCpp instead.")
        exit(1)
```

PR #659, opened 2023-06-07
Skip for empty query

doskoi:
Directly press ENTER or search space will get a meaningless result.


imartinez (approved):
Simple and effective, thanks!

PR #560, opened 2023-05-30
fixed the the csv file reading issue

ravindraprasad75:
Original code was written by [GuySarkinsky](https://github.com/GuySarkinsky) in issue #448

PR #660, opened 2023-06-07
Improving performance for PDF loader

doskoi:
The PyMuPDF brings much better performance for PDF loading.

https://pymupdf.readthedocs.io/en/latest/about.html#performance
Here is a comparison article:
https://medium.com/social-impact-analytics/comparing-4-methods-for-pdf-text-extraction-in-python-fd34531034f

It's about **5x-30x faster** for text extraction.

And In my few test cases, PyMuPDF have better result in CJK characters, Although they both may display a lot of garbled characters.

The different is PDFMinerLoader make a single document object but PyMuPDFLoader make each page as a document object.

This PR already contains the change that I've made for this update:
https://github.com/imartinez/privateGPT/pull/560/files

* Remove read subscript [0] only for _loader.load()_
* Update _results.append(doc)_ to  _results.extend(docs)_

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -9,7 +9,7 @@ from tqdm import tqdm
 from langchain.document_loaders import (
     CSVLoader,
     EverNoteLoader,
-    PDFMinerLoader,
+    PyMuPDFLoader,
     TextLoader,
     UnstructuredEmailLoader,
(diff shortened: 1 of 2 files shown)


imartinez (approved):
Great contribution, thanks!

PR #675, opened 2023-06-10
Update README.md instructions of .env file

nb-programmer:
Clarified to create a copy of example.env instead of renaming it to prevent accidentally removing from repo


imartinez (approved):
Thanks for the clarification! Will definitely help newcomers.

Issue #710, opened 2023-06-13
Update to allow a change in the number of threads available for use.


FearTheBadger:
**Is your feature request related to a problem? Please describe.**
I have more than the default number of threads (4) available for processing. I'd like to be able to use those extra threads.

**Describe the solution you'd like**
Add `n_threads` to the `llm` variable.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

TahirAhmadov:
I second this. The query response time is way too long and I could see that most of the cores are idle except for a few.

PR #711, opened 2023-06-13
Update to include `n_threads` as a modifiable variable.

FearTheBadger:
This should cover:

https://github.com/imartinez/privateGPT/issues/710


HarikrishnareddGali:
can you add a variable to use all the cores in cpu?

FearTheBadger:
@HarikrishnareddGali, this update will do that. A CPU has cores, a core has threads.

If you have one CPU with 4 cores, and each core has 4 threads, you can hypothetically update `n_threads` to 16 to utilize all 4 cores and 100% of the CPU.

PR #715, opened 2023-06-14
speedup: use all available CPU cores automatically in privateGPT.py

vnevoa:
Here's my proposal for using all available CPU cores automatically in privateGPT.py.
I have only used it with GPT4ALL, haven't tried LLAMA model.

### DIFF
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -35,7 +35,7 @@ def main():
         case "LlamaCpp":
             llm = LlamaCpp(model_path=model_path, n_ctx=model_n_ctx, n_batch=model_n_batch, callbacks=callbacks, verbose=False)
         case "GPT4All":
-            llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
+            llm = GPT4All(model=model_path, n_threads=len(os.sched_getaffinity(0)), n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
         case _default:
             print(f"Model {model_type} not supported!")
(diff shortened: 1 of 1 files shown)


privateGPT.py
```diff
@@ -35,7 +35,7 @@ def main():
         case "LlamaCpp":
             llm = LlamaCpp(model_path=model_path, n_ctx=model_n_ctx, n_batch=model_n_batch, callbacks=callbacks, verbose=False)
         case "GPT4All":
-            llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
+            llm = GPT4All(model=model_path, n_threads=len(os.sched_getaffinity(0)), n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
```
sime2408, line 38: `os.sched_getaffinity(pid)` and its counterpart `os.sched_setaffinity(pid, mask)` for setting the CPU affinity mask are only available on some Unix platforms. On platforms that do not support the `sched_setaffinity()` and `sched_getaffinity()` system calls, importing the os module will not define these functions. I, then propose not to use it and never try to exhaust 100% of the processor. Rather go with 80%:

```python
# Number of CPU cores
n_cpu = multiprocessing.cpu_count()

# Now, if you want to use, say, 80% of your CPU cores:
cpu_model_n_threads = int(0.8 * n_cpu)
```

rajeshsuramalla:
When I use "**n_threads=len(os.sched_getaffinity(0))**" throws me following error 
AttributeError: module 'os' has no attribute 'sched_getaffinity'.


"**n_threads=os.cpu_count()**" is worked for me.

PR #727, opened 2023-06-16
raise exception instead of print & exit

djm93dev:
It's better to raise an exception to handle the error more gracefully than abruptly exiting the script.


imartinez (approved):
Can't agree more. Thanks!

PR #852, opened 2023-07-13
Update privateGPT.py to work with python 3.8

shweshi:
**Error:** With python 3.8 we get error:
```
    match model_type:
          ^
SyntaxError: invalid syntax
```

**Reason:** The match/case syntax, otherwise known as Structural Pattern Matching, was only introduced in Python version 3.10.

**Fix:** Updated the code to not use match and use if/else instead.

Screenshot 2023-07-13 145651

### DIFF
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -36,7 +36,7 @@ def main():
     elif model_type == "GPT4All":
             llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', callbacks=callbacks, verbose=False)
     else:
-            print(f"Model {model_type} not supported!")
+            print(f"Model type {model_type} is not supported. Please choose one of the following: LlamaCpp, GPT4All")
             exit
         
     qa = RetrievalQA.from_chain_type(llm=llm, chain_type="stuff", retriever=retriever, return_source_documents= not args.hide_source)
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -31,14 +31,13 @@ def main():
     # activate/deactivate the streaming StdOut callback for LLMs
     callbacks = [] if args.mute_stream else [StreamingStdOutCallbackHandler()]
     # Prepare the LLM
-    match model_type:
-        case "LlamaCpp":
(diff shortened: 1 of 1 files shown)


imartinez (commented):
Thanks for the contribution. 
You've added changes beyond the scope of the PR. For example you've removed the usage of n_batch in the models, and moved from raise Exception to print and exit.

AITEK-DEV (changes requested):
# This code checks if the model type is "LlamaCpp" or "GPT4All".
# If not, it raises an error message.

if model_type == "LlamaCpp":
    # Creates a LlamaCpp model object.
    llm = LlamaCpp(model_path=model_path, n_ctx=model_n_ctx, n_batch=model_n_batch, callbacks=callbacks, verbose=False)
elif model_type == "GPT4All":
    # Creates a GPT4All model object.
    llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
else:
    # Raises an error message if the model type is not supported.
    raise Exception(f"Model type {model_type} is not supported. Please choose one of the following: LlamaCpp, GPT4All")

privateGPT.py
```diff
@@ -31,14 +31,13 @@ def main():
...
-            llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', n_batch=model_n_batch, callbacks=callbacks, verbose=False)
-        case _default:
-            # raise exception if model_type is not supported
-            raise Exception(f"Model type {model_type} is not supported. Please choose one of the following: LlamaCpp, GPT4All")
+    if model_type == "LlamaCpp":
+            llm = LlamaCpp(model_path=model_path, n_ctx=model_n_ctx, callbacks=callbacks, verbose=False)
+    elif model_type == "GPT4All":
+            llm = GPT4All(model=model_path, n_ctx=model_n_ctx, backend='gptj', callbacks=callbacks, verbose=False)
+    else:
+            print(f"Model type {model_type} is not supported. Please choose one of the following: LlamaCpp, GPT4All")
```
Simon-XG-Yan, line 39: Suggests `raise ValueError(f"{model_type} is not among LlamaCpp and GPT4All and not supported.")`

PR #950, opened 2023-08-14
Fixing Llamacpp context size bug

afonsoguerra:
I have noticed that changing MODEL_N_CTX wasn't changing the way the models were loaded in llamacpp. This is particularly obvious now with the larger llama2 context sizes. After some digging I found that this was down to the parameter being passed to langchain. That is what this pull request fixes. I have tested it and it is working as it should for llama2-chat models.

### DIFF
--- a/privateGPT.py
+++ b/privateGPT.py
@@ -33,7 +33,7 @@ def main():
     # Prepare the LLM
     match model_type:
         case "LlamaCpp":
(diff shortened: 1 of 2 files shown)


afonsoguerra:
Note, it might need both max_tokens and n_ctx, but if just max_tokens used if frequently crashes with a message similar to:
https://github.com/langchain-ai/langchain/issues/3751

Shivamdwivedi1403 (approved):

PR #957, opened 2023-08-16
Adding support to ingest files with extensions in uppercase

parampavar:
Files in the source_directory where ignored if their extensions where in uppercase like (*.PDF).
This change supports ingestion of files that match either lowercase or uppercase extensions like *.pdf or *.PDF. 
This can be enhanced further to support camelcase like *.Pdf at a later stage. The assumption is that this scenario is probably less than 5%.

### DIFF
--- a/ingest.py
+++ b/ingest.py
@@ -82,7 +82,7 @@ LOADER_MAPPING = {
 
 
 def load_single_document(file_path: str) -> List[Document]:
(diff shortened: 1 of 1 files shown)


imartinez (approved):

pascalrheaume:
The case sensitive extension problem to determine the file reader (e.g. .PDF file read with StringIterableReader) was not retrieved from the primordial branch fix.  Thanks!

PR #960, opened 2023-08-18
Add_sentence_transformers_2.2.2_to_required

TheCubicleJockey:
Adding sentence_transformers to requirements.txt. Two different computers had an issue and pip told me to install this in order to resolve the problem. Fix worked for both computers on python3


ThomThio:
Can confirm. Error only occurs when running privateGPT.py

imartinez (approved):

imartinez:
Thanks for the contribution!

Issue #990, opened 2023-08-29
Cannot submit more than 166 embeddings at once while ingesting


FelixHanusch:
**Describe the bug and how to reproduce it**
I pulled the repository and followed the steps described in the README. After running `python ingest.py` I am getting an error. Here is my complete output that gets printed to the console:
```
Creating new vectorstore
Loading documents from source_documents
Loading new documents: 100%|██████████████████████| 2/2 [00:05<00:00,  2.77s/it]
Loaded 2 new documents from source_documents
Split into 182 chunks of text (max. 500 tokens each)
Creating embeddings. May take some minutes...
Traceback (most recent call last):
  File "C:\Users\user\privateGPT\ingest.py", line 169, in <module>
    main()
  File "C:\Users\user\privateGPT\ingest.py", line 161, in main
    db = Chroma.from_documents(texts, embeddings, persist_directory=persist_directory, client_settings=CHROMA_SETTINGS, client=chroma_client)
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\langchain\vectorstores\chroma.py", line 612, in from_documents
    return cls.from_texts(
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\langchain\vectorstores\chroma.py", line 576, in from_texts
    chroma_collection.add_texts(texts=texts, metadatas=metadatas, ids=ids)
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\langchain\vectorstores\chroma.py", line 222, in add_texts
    raise e
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\langchain\vectorstores\chroma.py", line 208, in add_texts
    self._collection.upsert(
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\chromadb\api\models\Collection.py", line 298, in upsert
    self._client._upsert(
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\chromadb\api\segment.py", line 290, in _upsert
    self._producer.submit_embeddings(coll["topic"], records_to_submit)
  File "C:\Users\user\privateGPT\.venv\lib\site-packages\chromadb\db\mixins\embeddings_queue.py", line 127, in submit_embeddings
    raise ValueError(
ValueError:
                Cannot submit more than 166 embeddings at once.
                Please submit your embeddings in batches of size
                166 or less.
```

**Environment:**
 - OS / hardware: Windows 10 (MinGW is installed)
 - Python version: 3.10.2

[...]

PR #999, opened 2023-08-31
Batch embeddings to be processed by chromadb

imartinez:
Instead of adding all documents to ChromaDB at once, batch them to not go beyond the limit that makes ingestion fail.

Max batch size was introduced to ChromaDB in this PR https://github.com/chroma-core/chroma/pull/995

--

Fixes https://github.com/imartinez/privateGPT/issues/990


genglinxiao:
This PR should be at the chromadb side. no?

PR #1069, opened 2023-10-13
Colorized output

ReneNyffenegger:


ReneNyffenegger:
This pull request adds the Python library colorama to colorize the output created by privateGPT.py in the console.

requirements.txt
```diff
@@ -12,3 +12,4 @@ pandoc==2.3
 pypandoc==1.11
 tqdm==4.66.1
 sentence_transformers==2.2.2
+colorama>=0.4.3
```
lopagela, line 15: This should also be in the packaging system - i.e. `pyproject.toml`.
Also, maybe you want to add this as an optional dependency.

---
One should maybe also use the design pattern "Strategy" to decouple this feature being added while still preserving the code's readability.

Be30n (commented):
Ben

PR #1133, opened 2023-10-29
Drop loguru and use builtin `logging`

lopagela:
* Changed the 2 existing `print` in the `private_gpt` code base into actual python logging, stop using loguru (dependency will be dropped in a later commit).
* Try to use the `key=value` logging convention in logs (to indicate what dynamic values represents, and what is dynamic vs not).
* Using `%s` log style, so that the string formatting is pushed inside the logger, giving the ability to the logger to determine if the string need to be formatted or not (i.e. strings from debug logs might not be formatted if the log level is not debug)
* The (basic) builtin log configuration have been placed in `private_gpt/__init__.py` in order to initialize the logging system even before we start to launch any python code in `private_gpt` package (ensuring we get any initialization log formatted as we want to)
* Disabled `uvicorn` custom logging format, resulting in having uvicorn logs being outputted in our formatted.

Some more concise format could be used if we want to, especially:
```
COMPACT_LOG_FORMAT = '%(asctime)s.%(msecs)03d [%(levelname)s] %(name)s - %(message)s'
```

Python documentation and cookbook on logging for reference:
* https://docs.python.org/3/library/logging.html
* https://docs.python.org/3/howto/logging.html

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1917,24 +1917,6 @@ typing-inspect = ">=0.8.0"
 urllib3 = "<2"
 
 [[package]]
-name = "loguru"
-version = "0.7.2"
-description = "Python logging made (stupidly) simple"
-optional = false
-python-versions = ">=3.5"
-files = [
-    {file = "loguru-0.7.2-py3-none-any.whl", hash = "sha256:003d71e3d3ed35f0f8984898359d65b79e5b21943f78af86aa5491210429b8eb"},
-    {file = "loguru-0.7.2.tar.gz", hash = "sha256:e671a53522515f34fd406340ee968cb9ecafbc4b36c679da03c18fd8d0bd51ac"},
-]
-
-[package.dependencies]
-colorama = {version = ">=0.3.4", markers = "sys_platform == \"win32\""}
-win32-setctime = {version = ">=1.0.0", markers = "sys_platform == \"win32\""}
-
-[package.extras]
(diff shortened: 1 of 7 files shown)


lopagela:
Results after this PR: 
Screenshot 2023-10-29 at 17 07 34

When enabling `DEBUG` level:
Screenshot 2023-10-28 at 17 14 17

poetry.lock
```diff
@@ -5154,20 +5136,6 @@ files = [
 [package.extras]
 test = ["pytest (>=6.0.0)", "setuptools (>=65)"]
 
-[[package]]
```
lopagela, line 5157: Highlighting this one -- It has been removed as a side effect of `poetry remove loguru`

---

How should `poetry.lock` be managed? Should it be generated by some CICD tool? Or maintained by core dev of this project?

private_gpt/components/llm/custom/sagemaker.py
```diff
@@ -26,6 +27,8 @@
         CompletionResponseGen,
     )
 
+log = logging.getLogger(__name__)
```
pabloogc, line 30: tiny NIT but could you rename this to `logger` (and in the other file too)? 

Will be more consistent naming wise for the future.

pabloogc (approved):

private_gpt/components/llm/custom/sagemaker.py
lopagela: Sure sorry - I'm used to Java's Lombok `@Log4j2` that injects `log` variable into classes 😅 

I personally preferring `log` as it's shorter than `logger` (it makes lines shorter and seems more readable to me), but no problem, changing to `logger`

poetry.lock
pabloogc: I think we should ignore for the most part, it's just for consistent builds

pabloogc (approved):

PR #1144, opened 2023-10-31
Disable chromaDB anonymous information collection

Deeds67:
When running privateGPT, this was spotted in the logs:

```
chromadb.telemetry.posthog - Anonymized telemetry enabled. See https://docs.trychroma.com/telemetry for more information.
```

One can either disable it in the config object, as done in the PR, or as an env var `ANONYMIZED_TELEMETRY=False`

### DIFF
--- a/private_gpt/components/vector_store/vector_store_component.py
+++ b/private_gpt/components/vector_store/vector_store_component.py
@@ -1,6 +1,7 @@
 import typing
 
 import chromadb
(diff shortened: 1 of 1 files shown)


Deeds67:
Done

imartinez (approved):
Thank you, forgot to add this setting. We had it in the primordial branch.

PR #1163, opened 2023-11-04
Endpoint to delete documents ingested

lopagela:
A file that is ingested will be transformed into several documents (that are organized into nodes).
This endpoint is deleting documents (bits of a file). These bits can be retrieved thanks to the endpoint to list all the documents.

Also added logs in the scope of this PR

### DIFF
--- a/private_gpt/__main__.py
+++ b/private_gpt/__main__.py
@@ -5,8 +5,7 @@ import uvicorn
 from private_gpt.main import app
 from private_gpt.settings.settings import settings
 
-if __name__ == "__main__":
-    # Set log_config=None to do not use the uvicorn logging configuration, and
-    # use ours instead. For reference, see below:
-    # https://github.com/tiangolo/fastapi/discussions/7457#discussioncomment-5141108
-    uvicorn.run(app, host="0.0.0.0", port=settings.server.port, log_config=None)
+# Set log_config=None to do not use the uvicorn logging configuration, and
+# use ours instead. For reference, see below:
+# https://github.com/tiangolo/fastapi/discussions/7457#discussioncomment-5141108
+uvicorn.run(app, host="0.0.0.0", port=settings.server.port, log_config=None)
--- a/private_gpt/__main__.py
+++ b/private_gpt/__main__.py
@@ -5,7 +5,8 @@ import uvicorn
 from private_gpt.main import app
 from private_gpt.settings.settings import settings
 
-# Set log_config=None to do not use the uvicorn logging configuration, and
-# use ours instead. For reference, see below:
-# https://github.com/tiangolo/fastapi/discussions/7457#discussioncomment-5141108
-uvicorn.run(app, host="0.0.0.0", port=settings.server.port, log_config=None)
+if __name__ == "__main__":
+    # Set log_config=None to do not use the uvicorn logging configuration, and
+    # use ours instead. For reference, see below:
+    # https://github.com/tiangolo/fastapi/discussions/7457#discussioncomment-5141108
+    uvicorn.run(app, host="0.0.0.0", port=settings.server.port, log_config=None)
--- a/private_gpt/server/ingest/ingest_router.py
+++ b/private_gpt/server/ingest/ingest_router.py
@@ -53,9 +53,8 @@ def list_ingested() -> IngestResponse:
 def delete_ingested(doc_id: str) -> None:
     """Delete the specified ingested Document.
 
-    The `doc_id` can be obtained from the `GET /ingest/list` endpoint
-    The document will be effectively deleted from the document store (i.e.
-    from the directory specified in your configuration)
+    The `doc_id` can be obtained from the `GET /ingest/list` endpoint.
+    The document will be effectively deleted from your storage context.
     """
     service = root_injector.get(IngestService)
     service.delete(doc_id)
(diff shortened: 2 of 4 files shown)


private_gpt/server/ingest/ingest_service.py
```diff
@@ -153,7 +159,20 @@ def list_ingested(self) -> list[IngestedDoc]:
...
         return ingested_docs
+
+    def delete(self, doc_id: str) -> None:
+        """Delete an ingested document.
+
+        :raises ValueError: if the document does not exist
+        """
+        logger.info("Deleting the ingested document=%s in the doc store", doc_id)
+        self.storage_context.docstore.delete_ref_doc(doc_id)
+        # FIXME the documents are only deleted, and not in the vector store
```
imartinez, line 175: Here is the way to fix it. 
LlamaIndex has a built-in function to delete docs from the index (including the index_store, vector_store and doc_store). 

Documentation: https://docs.llamaindex.ai/en/stable/module_guides/indexing/document_management.html#deletion

The working code is (you'll need to import load_index_from_storage from llama_index, and store self.vector_store = vector_store_component.vector_store in the IngestService constructor)

```
def delete(self, doc_id: str) -> None:
        """Delete an ingested document.

        :raises ValueError: if the document does not exist
        """
        logger.info("Deleting the ingested document=%s in the doc store", doc_id)

        index = load_index_from_storage(self.storage_context, store_nodes_override=True)
        index.delete_ref_doc(doc_id, delete_from_docstore=True)

        # persist the index and nodes
        self.storage_context.persist(persist_dir=local_data_path)
```

There are two subtle details in the code to make it work.
store_nodes_override=True -> this is key to force the deletion update the doc_store and index_store
self.storage_context.persist() -> calling this is key to persist the changes

private_gpt/server/ingest/ingest_router.py
```diff
@@ -47,3 +47,21 @@ def list_ingested() -> IngestResponse:
...
+    """Delete the specified ingested Document.
+
+    The `doc_id` can be obtained from the `GET /ingest/list` endpoint
+    The document will be effectively deleted from the document store (i.e.
+    from the directory specified in your configuration)
+    """
+    service = root_injector.get(IngestService)
+    try:
+        service.delete(doc_id)
+    except ValueError as err:
```
imartinez, line 63: I'm no big fan of this. It is masking any error that may happen internally, not only "not found" errors. I'd let it crash, specially in this early stage of the project. So my proposal is to remove the try/catch altogether.

imartinez (changes requested):

private_gpt/server/ingest/ingest_service.py
lopagela: Adapted, thanks a lot 🙏

private_gpt/server/ingest/ingest_router.py
```diff
@@ -47,3 +47,15 @@ def list_ingested() -> IngestResponse:
     service = root_injector.get(IngestService)
     ingested_documents = service.list_ingested()
     return IngestResponse(object="list", model="private-gpt", data=ingested_documents)
+
+
+@ingest_router.delete("/ingest/{doc_id}", tags=["Ingestion"])
+def delete_ingested(doc_id: str) -> None:
+    """Delete the specified ingested Document.
+
+    The `doc_id` can be obtained from the `GET /ingest/list` endpoint
```
imartinez, line 56: This comment makes direct reference to the "local" setup, but PrivateGPT can be setup to work with different storage options (like remote vector and structured databases). I'd suggest to update this comment making it generic and talking about directories. Also, the doc is deleted not only from the document store. Let's make this more generic (check the rest of API docs)

imartinez (changes requested):
Last minor changes. Thanks!

private_gpt/__main__.py
```diff
@@ -5,7 +5,8 @@
 from private_gpt.main import app
 from private_gpt.settings.settings import settings
 
-# Set log_config=None to do not use the uvicorn logging configuration, and
-# use ours instead. For reference, see below:
-# https://github.com/tiangolo/fastapi/discussions/7457#discussioncomment-5141108
-uvicorn.run(app, host="0.0.0.0", port=settings.server.port, log_config=None)
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
lopagela: Only by convenience on my end (to click on the little ⏯️ button in my PyCharm) - sorry, will remove

On another note, for `__main__.py`, this block will always be evaluated, given that `__name__` will always return `__main__`, per the filename.

Otherwise, `if __name__ == "__main__":` is the common pattern to run code conditionally on how the python file get invoked (just like in `private_gpt/ui/ui.py`)

private_gpt/server/ingest/ingest_service.py
```diff
@@ -100,7 +105,9 @@ def ingest(self, file_name: str, file_data: AnyStr | Path) -> list[IngestedDoc]:
                     else:
                         path_to_tmp.write_text(str(file_data))
                     documents = reader.load_data(path_to_tmp)
-
+        logger.info(
+            "Transformed file=%s into count=%s documents", file_data, len(documents)
```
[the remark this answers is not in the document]
lopagela: You are 100% correct - my bad - must have been auto-completion error

imartinez (approved):

BennisonDevadoss:
While uploading a single document, I won't get only one doc_id, ingest post method would return multiple doc_id based on the document size, In this case, should we call the delete API multiple times to delete the uploaded file?

lopagela:
> While uploading a single document, I won't get only one doc_id, ingest post method would return multiple doc_id based on the document size, In this case, should we call the delete API multiple times to delete the uploaded file?

Yes, the client can `for` loop on the IDs that are returned by the list endpoint

PR #1165, opened 2023-11-05
Disable Gradio Analytics

lopagela:
Gradio analytics can be disabled by either using the kwargs `enable_analytics` on `gr.Blocks`, or by setting the env variable `GRADIO_ANALYTICS_ENABLED` to something different from `True`.

Since that Gradio does not seem to respect their code contract (around `enable_analytics`), and that they are performing other operations only based on the value of `GRADIO_ANALYTICS_ENABLED` (c.f. `gradio.strings` https://github.com/gradio-app/gradio/blob/main/gradio/strings.py#L39), we are disabling gradio analytics by setting the required env variable to `False`.

Note: Setting an environment variables using `os.environ['foo'] = 'bar'` on system that are not based on unix might not work.

c.f. https://docs.python.org/3/library/os.html#os.environ for details on how `os.environ` works and all its caveats

See https://github.com/imartinez/privateGPT/pull/1144 for the similar PR on chromaDB telemetry (that motivated this PR)

### DIFF
--- a/private_gpt/__init__.py
+++ b/private_gpt/__init__.py
@@ -1,11 +1,23 @@
 """private-gpt."""
 import logging
+import os
 
 # Set to 'DEBUG' to have extensive logging turned on, even for libraries
-ROOT_LOG_LEVEL = "INFO"
+ROOT_LOG_LEVEL = "DEBUG"
 
 PRETTY_LOG_FORMAT = (
     "%(asctime)s.%(msecs)03d [%(levelname)-8s] %(name)+25s - %(message)s"
 )
 logging.basicConfig(level=ROOT_LOG_LEVEL, format=PRETTY_LOG_FORMAT, datefmt="%H:%M:%S")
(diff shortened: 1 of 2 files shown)


lopagela:
After this change, there will be **absolutely no** HTTP call made to gradio API, cleaning the `DEBUG` log, and preserving anonymity of users

imartinez (dismissed):

private_gpt/__init__.py
```diff
@@ -1,11 +1,23 @@
 """private-gpt."""
 import logging
+import os
 
 # Set to 'DEBUG' to have extensive logging turned on, even for libraries
-ROOT_LOG_LEVEL = "INFO"
+ROOT_LOG_LEVEL = "DEBUG"
```
lopagela, line 6:
```suggestion
ROOT_LOG_LEVEL = "INFO"
```

imartinez (approved):

Issue #1166, opened 2023-11-05
Error: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte


yaziciali:
20:56:50.251 [ERROR   ]                  __main__ - Failed to ingest document: SRC/outlook/Data/Messages/94/5E6F4873-D757-4141-8EF2-D6AC865040A7.olk15Message. Error: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte
20:56:50.251 [INFO    ]                  __main__ - Document 18875 of 239575 (7.88%)
20:56:50.251 [INFO    ]                  __main__ - Started ingesting SRC/outlook/Data/Messages/94/5E2758E1-8DAC-4F73-BFF6-6CA994ED8F23.olk15Message
^CTraceback (most recent call last):
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 83, in <module>
    _recursive_ingest_folder(path)
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 63, in _recursive_ingest_folder
    _recursive_ingest_folder(file_path)
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 63, in _recursive_ingest_folder
    _recursive_ingest_folder(file_path)
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 63, in _recursive_ingest_folder
    _recursive_ingest_folder(file_path)
  [Previous line repeated 1 more time]
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 61, in _recursive_ingest_folder
    _do_ingest(file_path)
  File "/Users/ayazici/privateGPT/scripts/ingest_folder.py", line 70, in _do_ingest
    ingest_service.ingest(changed_path.name, changed_path)
  File "/Users/ayazici/privateGPT/private_gpt/server/ingest/ingest_service.py", line 81, in ingest
    text = file_data.read_text()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pathlib.py", line 1058, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1171, opened 2023-11-06
Update ingest_service.py to fix issue Error: 'utf-8' codec can't decode

yaziciali:
To fix issue: https://github.com/imartinez/privateGPT/issues/1166

Error: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte #1166

### DIFF
--- a/private_gpt/server/ingest/ingest_service.py
+++ b/private_gpt/server/ingest/ingest_service.py
@@ -1,4 +1,5 @@
 import tempfile
+import chardet # Chardet must be put in requirements or manually install with pip install chardet
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, AnyStr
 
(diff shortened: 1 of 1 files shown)


private_gpt/server/ingest/ingest_service.py
```diff
@@ -1,4 +1,5 @@
 import tempfile
+import chardet # Chardet must be put in requirements or manually install with pip install chardet
```
pabloogc, line 2: You should add the dependency,

`poetry add chardet`

private_gpt/server/ingest/ingest_service.py
```diff
@@ -77,7 +78,10 @@ def ingest(self, file_name: str, file_data: AnyStr | Path) -> list[IngestedDoc]:
             # Read as a plain text
             string_reader = StringIterableReader()
             if isinstance(file_data, Path):
-                text = file_data.read_text()
+                with open(file_data, 'rb') as f2:
```
pabloogc, line 81: Try to give this variables proper names. Maybe `file_handle` instead of `f2` and `charset` instead of `result2`

pabloogc (changes requested):

pabloogc:
Make sure to run the formatter and lint `make check`

private_gpt/server/ingest/ingest_service.py
lopagela: I would go even further, and could you also try to avoid to read the file in it's entirety to detect the charset?

More information in `chardet` documentation: https://chardet.readthedocs.io/en/latest/usage.html#advanced-usage

While the existing implementation work, it is suboptimal in the sense that it reads the file in its entirety before re-reading it?

One could also read the file in binary mode, store it in `text_bin`, and then run `chardet` on it, so that we have `text = text_bin.decode(encoding=chardet.detect(text_bin))`

This is doing only a single read, and it re-uses the buffer in memory instead of re-reading it.

PR #1176, opened 2023-11-07
Parse JSON files using llama_index JSONReader

lopagela:
Patch the default list of llama_index to support JSON files. This injection of JSON documents should improve the comprehension in JSON files, as there is a parsing of JSON files.

It **might** avoid the docstore pollution that we've been seeing in reported issues lately

### DIFF
--- a/private_gpt/server/ingest/ingest_service.py
+++ b/private_gpt/server/ingest/ingest_service.py
@@ -14,9 +14,7 @@ from llama_index import (
     load_index_from_storage,
 )
 from llama_index.node_parser import SentenceWindowNodeParser
-from llama_index.readers.file.base import (
-    DEFAULT_FILE_READER_CLS as ORIGINAL_DEFAULT_FILE_READER_CLS,
-)
+from llama_index.readers.file.base import DEFAULT_FILE_READER_CLS
 from pydantic import BaseModel, Field
 
 from private_gpt.components.embedding.embedding_component import EmbeddingComponent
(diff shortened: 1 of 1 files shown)


private_gpt/server/ingest/ingest_service.py
```diff
@@ -6,14 +6,17 @@
...
     ServiceContext,
     StorageContext,
     StringIterableReader,
     VectorStoreIndex,
     load_index_from_storage,
 )
 from llama_index.node_parser import SentenceWindowNodeParser
-from llama_index.readers.file.base import DEFAULT_FILE_READER_CLS
+from llama_index.readers.file.base import (
+    DEFAULT_FILE_READER_CLS as ORIGINAL_DEFAULT_FILE_READER_CLS,
```
imartinez, line 18: Small detail to make it more clear for future readers: instead of patching the default file reader we could:
- Keep on importing DEFAULT_FILE_READER_CLS as it was before
- Create our own FILE_READER_CLS var:
FILE_READER_CLS = DEFAULT_FILE_READER_CLS.copy()
FILE_READER_CLS.update(
    {
        ".json": JSONReader,
    }
)

- Use our FILE_READER_CLS in the ingest function body.

private_gpt/server/ingest/ingest_service.py
```diff
@@ -6,14 +6,17 @@
 from injector import inject, singleton
 from llama_index import (
     Document,
+    JSONReader,
```
imartinez, line 9: This may require adding an extra dependency to poetry. Did you check?

imartinez (changes requested):
Love it! Just a couple comments.

private_gpt/server/ingest/ingest_service.py
lopagela: I did by reading the implementation in `llama_index`: it leverage python's implementation of `json`, so no new dependency.

private_gpt/server/ingest/ingest_service.py
lopagela: Done

imartinez (approved):

PR #1191, opened 2023-11-09
Refactor UI state management

lopagela:
The class `PrivateGptUi` will hold the definition of the UI, as well as it's state.

The `PrivateGptUi` becomes stateless, and the UI will reload the documents that are ingested everytime it needs to (i.e. after a file upload, or every time the UI reloads)

----

Documentation used to change `gradio`'s behaving:

* `DataFrame`: https://www.gradio.app/docs/dataframe
* `UploadButton`: https://www.gradio.app/docs/uploadbutton
* `Blocks`: https://www.gradio.app/docs/blocks
* State in Blocks: https://www.gradio.app/guides/state-in-blocks

### DIFF
--- a/private_gpt/ui/ui.py
+++ b/private_gpt/ui/ui.py
@@ -1,4 +1,8 @@
+"""
+This file should be imported only and only if you want to run the UI locally.
+"""
 import itertools
+import logging
 from collections.abc import Iterable
 from pathlib import Path
 from typing import Any, TextIO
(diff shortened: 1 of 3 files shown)


private_gpt/ui/ui.py
lopagela, line 1: # Note to reviewers

One might better look at the commit changes individually to capture the logic of the changes made in this file.

The work was done this way:

1. Move the existing code into a class (no code changes in the core logic)
2. Remove the existing state, and instead, trigerring a every upload, and at every page refresh

scripts/ingest_folder.py
```diff
@@ -69,8 +69,10 @@ def _do_ingest(changed_path: Path) -> None:
             logger.info(f"Started ingesting {changed_path}")
             ingest_service.ingest(changed_path.name, changed_path)
             logger.info(f"Completed ingesting {changed_path}")
-    except Exception as e:
-        logger.error(f"Failed to ingest document: {changed_path}. Error: {e}")
+    except Exception:
+        logger.exception(
```
lopagela, line 73: `error` -> `exception` to display the stack trace when folks are ingesting files using this script.

It will be helpful to debug folks (that seems to have `llama_index` complaining about missing packages while it is installed

imartinez (approved):
Amazing! Thanks!

PR #1203, opened 2023-11-10
Add simple Basic auth

lopagela:
To enable the basic authentication, one must set `server.auth.enabled` to true.

The static string defined in `server.auth.secret` must be set in the header `Authorization`.

The health check endpoint will always be accessible, no matter the API auth configuration.

----

The `authenticated` "bean"/"method-in-the-chain" is declared as a dependency in the API router that requires `authenticated` connections. c.f. `faspAPI` documentation for details on `dependencies`: https://fastapi.tiangolo.com/tutorial/bigger-applications/

`Security` section of the fastAPI is also a good read: https://fastapi.tiangolo.com/tutorial/security/

### DIFF
--- a/private_gpt/server/utils/auth.py
+++ b/private_gpt/server/utils/auth.py
@@ -38,8 +38,6 @@ logger = logging.getLogger(__name__)
 
 def _simple_authentication(authorization: Annotated[str, Header()] = "") -> bool:
     """Check if the request is authenticated."""
-    if not settings.server.auth.enabled:
-        return True
     if not secrets.compare_digest(authorization, settings.server.auth.secret):
         # If the "Authorization" header is not the expected one, raise an exception.
         raise NOT_AUTHENTICATED
@@ -64,8 +62,7 @@ else:
         _simple_authentication: Annotated[bool, Depends(_simple_authentication)]
     ) -> bool:
         """Check if the request is authenticated."""
-        if not settings.server.auth.enabled:
-            return True
+        assert settings.server.auth.enabled
         if not _simple_authentication:
             raise NOT_AUTHENTICATED
         return True
--- a/private_gpt/server/utils/auth.py
+++ b/private_gpt/server/utils/auth.py
@@ -17,14 +17,15 @@ import secrets
 from typing import Annotated
 
 from fastapi import Depends, Header, HTTPException
+
 from private_gpt.settings.settings import settings
 
 # 401 signify that the request requires authentication.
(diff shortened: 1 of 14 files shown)


imartinez (dismissed):
Looks good to me! We me be overusing the 
`if not settings.server.auth.enabled:`
check in auth.py, but I think it makes it safer for future refactors.

As a general note, we should be documenting this (also the previous CORS support) in `docs/description.md` and then executing `make api-docs` to generate the udpated version of the api docs github pages, then commit those changes.

private_gpt/server/utils/auth.py
```diff
@@ -0,0 +1,71 @@
...
+
+else:
+    logger.info("Defining the given authentication mechanism for the API")
+
+    # Method to be used as a dependency to check if the request is authenticated.
+    def authenticated(
+        _simple_authentication: Annotated[bool, Depends(_simple_authentication)]
+    ) -> bool:
+        """Check if the request is authenticated."""
+        if not settings.server.auth.enabled:
```
pabloogc, line 67: do you need this check here again? it's also inside `_simple_authentication`

private_gpt/server/utils/auth.py
```diff
@@ -0,0 +1,71 @@
...
+NOT_AUTHENTICATED = HTTPException(
+    status_code=401,
+    detail="Not authenticated",
+    headers={"WWW-Authenticate": 'Basic realm="All the API", charset="UTF-8"'},
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _simple_authentication(authorization: Annotated[str, Header()] = "") -> bool:
```
pabloogc, line 39: Please add an integration test using the test client. Any endpoint will do since all use the same auth.

pabloogc (changes requested):

private_gpt/server/utils/auth.py
lopagela: Done - some method was used in the test because of the way the existing is made

I guess the tests can be refactored later (/improved)

private_gpt/server/utils/auth.py
lopagela: Removed

pabloogc (approved):

private_gpt/server/utils/auth.py
pabloogc: yeah, we were not careful with the setting and used them as a constant value and that is also making the tests annoying to write. I'll look into improving that.

PR #1206, opened 2023-11-11
Add sources to completions APIs and UI

imartinez:
Reuse `Chunk` object to append the sources used during completions to the completions response.
Affects both chat/completions and plain /completions

It has been integrated in the API and UI (see attachments).

Note 1: `Chunk`s returned as sources will always have their `previous_texts` and `next_texts` set to none, given those are intended to be only used in the /chunks API - therefore we could eventually separate both objects (Chunks used for source representation, and /chunks API response)

Note 2: this information is also being added to the stream API, making the streaming chunks too big in my opinion, potentially affecting performance; should we remove sources from the streaming API? 

Screenshot 2023-11-11 at 10 52 51

Screenshot 2023-11-11 at 10 53 28


imartinez:
On note 2, we could maybe add an input parameter to the call, same as "stream" that is "include_context_sources" or something like that.

lopagela (approved):
LGTM even though I'm not capable at 100% to validate what is happening 😅

private_gpt/server/chunks/chunks_service.py
```diff
@@ -24,17 +24,33 @@ class Chunk(BaseModel):
...
             [
                 "New leads came from Google Ads campaign.",
                 "The campaign was run by the Marketing Department",
             ]
-        ]
+        ],
     )
 
+    @classmethod
+    def from_node(cls: type["Chunk"], node: NodeWithScore) -> "Chunk":
```
pabloogc, line 41: you couldn't use the Chunk type here? circular reference?

private_gpt/open_ai/openai_models.py
```diff
@@ -48,8 +52,11 @@ class OpenAICompletion(BaseModel):
     choices: list[OpenAIChoice]
 
     @classmethod
-    def from_text(
-        cls, text: str | None, finish_reason: str | None = None
+    def from_text_and_sources(
```
pabloogc, line 55: tiny nit, but if you name the function `from_text_and_sources` I would expect the sources to be non mandatory, and `from_text` to still be there but with the sources optional

pabloogc (approved):

private_gpt/server/chunks/chunks_service.py
imartinez: The reason is that this class method is defined within Chunk class definition. So Chunk type is not ready.

private_gpt/open_ai/openai_models.py
imartinez: Done!

lopagela (approved):

PR #1209, opened 2023-11-11
Update poetry lock

lopagela:
Updated `llama-index` to `0.8.67` to avoid users to have issues around `openai` import.

Re-generated the poetry.lock file by running:

```bash
rm -rf .venv/ poetry.lock

python3 -m venv .venv && source .venv/bin/activate && pip install --upgrade pip poetry && poetry install --with ui,local
```

Tested locally, and the basic call through the UI works as expected.


settings.yaml
```diff
@@ -15,7 +15,7 @@ ui:
   path: /
 
 llm:
-  mode: mock
+  mode: local
```
lopagela, line 18: Noteworthy change

pabloogc (approved):

imartinez (approved):

pabloogc:
looks like the bump to llama Index broke a type

lopagela:
> looks like the bump to llama index broke a type

@pabloogc , indeed, thank you, will fix (this is a warning that have been transformed into an error apparently :sweat_smile: -- `enum` in `Field` is not authorized anymore)

Fixing

lopagela:
@pabloogc fixed 

As explained in the commit message, I had to fight with function signature and do one extra `cast`

pabloogc (approved):

Issue #1189, opened 2023-11-09
page label error


iplayfast:
Uploaded some cpp files as a doc.
Asked it to list the functions in the file.
The screen showed a "page label error" and "error" but otherwise did nothing.
At the console side
Traceback (most recent call last):
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/queueing.py", line 407, in call_prediction
    output = await route_utils.call_process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/route_utils.py", line 226, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/blocks.py", line 1550, in process_api
    result = await self.call_function(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/blocks.py", line 1199, in call_function
    prediction = await utils.async_iteration(iterator)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/utils.py", line 519, in async_iteration
    return await iterator.__anext__()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/utils.py", line 623, in asyncgen_wrapper
    async for response in f(*args, **kwargs):
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/chat_interface.py", line 437, in _stream_fn
    first_response = await async_iteration(generator)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/chris/anaconda3/envs/privategpt/lib/python3.11/site-packages/gradio/utils.py", line 519, in async_iteration
    return await iterator.__anext__()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1212, opened 2023-11-12
Curate sources to avoid the UI crashing

imartinez:
Remove sources from chat history to avoid confusing the LLM

Fixes https://github.com/imartinez/privateGPT/issues/1189

### DIFF
--- a/private_gpt/ui/ui.py
+++ b/private_gpt/ui/ui.py
@@ -9,18 +9,28 @@ import gradio as gr  # type: ignore
 from fastapi import FastAPI
 from gradio.themes.utils.colors import slate  # type: ignore
 from llama_index.llms import ChatMessage, ChatResponse, MessageRole
+from pydantic import BaseModel
 
 from private_gpt.di import root_injector
 from private_gpt.server.chat.chat_service import ChatService, CompletionGen
-from private_gpt.server.chunks.chunks_service import ChunksService
+from private_gpt.server.chunks.chunks_service import Chunk, ChunksService
 from private_gpt.server.ingest.ingest_service import IngestService
 from private_gpt.settings.settings import settings
 from private_gpt.ui.images import logo_svg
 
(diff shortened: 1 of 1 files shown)


private_gpt/ui/ui.py
```diff
@@ -32,6 +42,23 @@ def __init__(self) -> None:
...
+                and "file_name" in chunk.document.doc_metadata
+                else "-",
+                page=chunk.document.doc_metadata["page_label"]
+                if chunk.document.doc_metadata
+                and "page_label" in chunk.document.doc_metadata
+                else "-",
+                text=chunk.text,
+            )
+            for chunk in sources
+        }
```
lopagela, line 60: Could have been in `Source` class

```python
# Usage would look like
Source.curated_from_chunks(sources)
```

lopagela: There are two identical check made on `bool(chunk.document.doc_metadata)` - could be reduced to one

Additionally, accessing to `page_label` and `file_name` can still cause key errors (the calls to the dict should be made with `.get(key, _default_value)`

private_gpt/ui/ui.py
```diff
@@ -9,18 +9,28 @@
...
 
-
 UI_TAB_TITLE = "My Private GPT"
+SOURCES_SEPARATOR = "\n\n Sources: \n"
+
+
+class Source(BaseModel):
+    file: str
+    page: str
+    text: str
```
lopagela, line 30: Nice idea to have that dedicated object, it reduce the code dup around metadata management ✌️

lopagela (commented):
I believe the key not found error is still present

private_gpt/ui/ui.py
imartinez: Moved and rewrote the whole method addressing these two comments. Thanks!

pabloogc (approved):

Issue #1210, opened 2023-11-12
Internal Server Error when calling DELETE on /v1/ingest/{doc_id} endpoint


frankwmoyer:
I deployed privateGPT locally without connection to OpenAI.

Here are the results of the `/v1/ingest/list` endpoint:

```
[...]

PR #1220, opened 2023-11-12
Reuse existing stored index during ingestion

imartinez:
So far PrivateGPT has been creating one index per ingested file. 
That makes it difficult to load the index from disk to perform update or delete operations, and it was causing errors during deletion. 

We now reuse the existing index if any. That way we ensure a single index for all files ingested, making it easier to operate with it.

### DIFF
--- a/private_gpt/server/ingest/ingest_service.py
+++ b/private_gpt/server/ingest/ingest_service.py
@@ -133,14 +133,27 @@ class IngestService:
             document.excluded_embed_metadata_keys = ["doc_id"]
             # We don't want the LLM to receive these metadata in the context
             document.excluded_llm_metadata_keys = ["file_name", "doc_id", "page_label"]
-        # create vectorStore index
-        VectorStoreIndex.from_documents(
-            documents,
-            storage_context=self.storage_context,
-            service_context=self.ingest_service_context,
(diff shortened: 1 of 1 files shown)


imartinez:
Fixes https://github.com/imartinez/privateGPT/issues/1210

pabloogc (approved):

private_gpt/server/ingest/ingest_service.py
```diff
@@ -133,14 +133,27 @@ def _save_docs(self, documents: list[Document]) -> list[IngestedDoc]:
...
+                storage_context=self.storage_context,
+                service_context=self.ingest_service_context,
+                store_nodes_override=True,  # Force store nodes in index and document stores
+                show_progress=True,
+            )
+            for doc in documents:
+                index.insert(doc)
+        except ValueError:
+            # Or create a new one if there is none
+            VectorStoreIndex.from_documents(
```
pabloogc, line 149: feels weird that you don't use the results of this line, everything done implicitly I guess when you pass a list of documents

pabloogc (approved):

private_gpt/server/ingest/ingest_service.py
imartinez: exactly...

PR #1228, opened 2023-11-13
feat: Qdrant support

Anush008:
This PR intends to add [Qdrant](https://qdrant.tech/) as a supported vector database.

The database to be used can be configured using the new `vectorstore.database` property. The default values are set, ensuring no breaking change in the current behaviour.


private_gpt/components/vector_store/vector_store_component.py
```diff
@@ -36,22 +40,58 @@ class VectorStoreComponent:
...
+                    VectorStore,
+                    BatchedChromaVectorStore(
+                        chroma_client=chroma_client, chroma_collection=chroma_collection
+                    ),
+                )
+
+            case "qdrant":
+                try:
+                    from llama_index.vector_stores.qdrant import QdrantVectorStore
+                    from qdrant_client import QdrantClient
```
imartinez, line 65: Given the installation of the package is optional, it won't be installed by the CI. 
The best solution I can think of is to make mypy ignore that import. 

```suggestion
                    from qdrant_client import QdrantClient  # type: ignore
```

imartinez (dismissed):
Amazing PR. Thanks for the clarity, the configurability and the detail to documentation. 

I tried it as QdrantLocal and works perfect (faster than Chromadb at around x2 in my first tests).

As a piece of unrelated feedback (definitely not to be fixed in the scope of this PR, but it will affect people adoption of qdrant), the configuration of QdrantClient is a little misleading as can be seen in the actual implementation of it:

```
 if location == ":memory:":
            self._client = QdrantLocal(location=location)
        else:
            if path is not None:
                self._client = QdrantLocal(location=path)
            else:
                if location is not None and url is None:
                    url = location
                self._client = QdrantRemote(
```

It is hard to understand how to configure it for local usage vs remote vs on memory.

Again, thanks a lot for the integration!

imartinez (approved):

PR #1249, opened 2023-11-15
feat: enable resume download for hf_hub_download

lonelam:
enable auto resume download by default to prevent always restart on downloading large file with poor network connection.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -1,10 +1,18 @@
 #!/usr/bin/env python3
 import os
+import argparse
 
 from huggingface_hub import hf_hub_download, snapshot_download
 
 from private_gpt.paths import models_path, models_cache_path
 from private_gpt.settings.settings import settings
+
+resume_download = True
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(prog='Setup: Download models from huggingface')
+    parser.add_argument('--resume', default=True, action=argparse.BooleanOptionalAction, help='Enable/Disable resume_download options to restart the download progress interrupted')
+    args = parser.parse_args()
+    resume_download = args.resume
 
 os.makedirs(models_path, exist_ok=True)
 embedding_path = models_path / "embedding"
@@ -24,6 +32,7 @@ hf_hub_download(
     filename=settings().local.llm_hf_model_file,
     cache_dir=models_cache_path,
     local_dir=models_path,
+    resume_download=resume_download,
 )
 
 print("LLM model downloaded!")


private_gpt/settings/settings.py
```diff
@@ -92,6 +92,7 @@ class LocalSettings(BaseModel):
     llm_hf_repo_id: str
     llm_hf_model_file: str
     embedding_hf_model_name: str
+    llm_hf_resume_download: bool
```
lopagela, line 95: Suggesting to make it default to `False` if no values are provided in documentation
```suggestion
    llm_hf_resume_download: bool = False
```

settings.yaml
```diff
@@ -30,6 +30,7 @@ local:
   llm_hf_repo_id: TheBloke/Mistral-7B-Instruct-v0.1-GGUF
   llm_hf_model_file: mistral-7b-instruct-v0.1.Q4_K_M.gguf
   embedding_hf_model_name: BAAI/bge-small-en-v1.5
+  llm_hf_resume_download: True
```
lopagela, line 33: Preferring json style boolean (even though `True` is [supported](https://yaml.org/type/bool.html)) as it stay consistent with the `false` in this file.

```suggestion
  llm_hf_resume_download: true
```

lopagela (commented):
Suggesting to make it `False` by default in `settings.py` to do not explicitly require it in the chain of `settings*.yaml`

pabloogc (changes requested):
I think going for a very simple solution here is way better. Just making it always resume the download seems like a healthy default.

That way you don't have to overcomplicate with a setting that realistically nobody will need to change.

lonelam:
@pabloogc , thank you for your review and suggestions. I agree that resuming downloads by default is a prudent approach. However, to address scenarios where resuming may not be ideal, such as when the model file in the HuggingFace repository has been updated, I propose adding an optional setting in the LocalSettings class. This would allow users to choose whether to resume downloads from the last breakpoint or start afresh, thereby ensuring flexibility and robustness in handling different situations.

pabloogc:
> @pabloogc , thank you for your review and suggestions. I agree that resuming downloads by default is a prudent approach. However, to address scenarios where resuming may not be ideal, such as when the model file in the HuggingFace repository has been updated, I propose adding an optional setting in the LocalSettings class. This would allow users to choose whether to resume downloads from the last breakpoint or start afresh, thereby ensuring flexibility and robustness in handling different situations.

I still wouldn't expose it as a setting, instead pass a parameter to the setup script `scripts/setup --resume=true/false` defaulting to true should be enough for 99.9% of scenarios.

private_gpt/settings/settings.py
lonelam: Has been removed from settings.

settings.yaml
lonelam: The option has been removed from settings.

pabloogc (approved):

PR #1253, opened 2023-11-16
Fix/qdrant support

imartinez:
There was an issue with local Qdrant (using on-disk sqlite) on different OS systems related to multi-thread access to the db.


lopagela (approved):

lopagela:
Flagging @Anush008 for visibility and potentially feedback on the solution

PR #1238, opened 2023-11-14
🌿 Introduce Fern Docs

dannysheridan:
This PR adds the open source, free tier of **Fern Docs** for PrivateGPT.

PR #1254, opened 2023-11-16
docs: Rearrange vectorstores section

Anush008:
Move the vectorstore section in the docs to a distinct spot and use heading H2(##) for the title for hyperlinking purposes.


imartinez:
We are reorganizing the documentation as a whole using Fern in the coming days 
https://github.com/imartinez/privateGPT/pull/1238

Anush008:
Saw the new docs. Currently we can't link to vectorstores section of the docs. Which is something we'd be needing to share on socials.

imartinez:
> Saw the new docs. Currently we can't link to the vectorstores section of the docs due to using #### for the heading. Which is something we'd be needing to share on socials.

Good point. Let me work on it before we announce it then. Thanks!

Issue #1227, opened 2023-11-13
API triggered document ingestion on Windows causes permission error


jaydavid:
When running locally on a Windows 11 machine, I am able to interact with the UI and upload files without issue.

However, when I attempt to connect to the server via another application using the API ingestion endpoint `/v1/ingest` I am seeing a permissions issue with the temporary file that's created to facilitate the ingestion. A snippet of the error:

```console
  File "C:\repos\privateGPT\private_gpt\server\ingest\ingest_router.py", line 38, in ingest
    ingested_documents = service.ingest(file.filename, file.file.read())
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\repos\privateGPT\private_gpt\server\ingest\ingest_service.py", line 118, in ingest
    path_to_tmp.write_bytes(file_data)
  File "C:\Users\Jay\AppData\Local\Programs\Python\Python311\Lib\pathlib.py", line 1067, in write_bytes
    with self.open(mode='wb') as f:
         ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Jay\AppData\Local\Programs\Python\Python311\Lib\pathlib.py", line 1044, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\Jay\\AppData\\Local\\Temp\\tmpxq8kqnc9'
```

Here is the demo code that I'm using to exercise the API. This is running on the node side of a next.js app.

```typescript
    let file = await fileRepo.get(fileId);
    let formData = new FormData();
    formData.append("file", base64toBlob(file.data), file.name);
    console.log("ingesting file...");
    const response = await fetch("http://127.0.0.1:8001/v1/ingest", {
      method: "POST",
      body: formData,
    });
```

I am able to correct this issue locally if I make the following change, which I believe is due to the way that python requisitions temporary files on a Windows machine. The change is made in the `/private_gpt/server/ingest/ingest_service.py` file.

Current:
```python
                with tempfile.NamedTemporaryFile() as tmp:
                    path_to_tmp = Path(tmp.name)
                    if isinstance(file_data, bytes):
                        path_to_tmp.write_bytes(file_data)
                    else:
[...]

PR #1260, opened 2023-11-17
fix: Windows 11 tmp file permission

pabloogc:
Fixes https://github.com/imartinez/privateGPT/issues/1227


danielgallegovico (approved):

imartinez (approved):

lopagela:
@pabloogc I don't think this is actually fixed

pabloogc:
> @pabloogc I don't think this is actually fixed

it's not, was worth the shot tho, will try with another API (os.close) see if that resolves it

PR #1265, opened 2023-11-17
Trigger `fern` jobs like before, and only when the content of `fern/` change

lopagela:
Trying to debug and understand why `fern` jobs does not kick in


imartinez:
Done in https://github.com/imartinez/privateGPT/pull/1273

PR #1274, opened 2023-11-18
🌿 introduce `x-fern-streaming` extension

dsinghvi:
In the `x-fern-streaming` extension you can specify a streaming version and non-streaming version of an endpoint. See the [docs](https://docs.buildwithfern.com/api-definition/open-api-specification/openapi/extensions#streaming)

### DIFF
--- a/fern/fern.config.json
+++ b/fern/fern.config.json
@@ -1,4 +1,4 @@
 {
   "organization": "privategpt",
-  "version": "0.15.0-rc80"
+  "version": "0.15.0-rc88"
 }
--- a/fern/openapi/openapi.json
+++ b/fern/openapi/openapi.json
@@ -35,6 +35,15 @@
             },
             "required": true
           },
+          "x-fern-streaming": {
+            "stream-condition": "$request.stream",
+            "response": {
+              "$ref": "#/components/schemas/OpenAICompletion"
+            },
+            "response-stream": {
+              "$ref": "#/components/schemas/OpenAICompletion"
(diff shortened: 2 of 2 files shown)


imartinez:
Instead of modifying openapi.json directly, we'd need to add the extension to FastAPI endpoint definition in both /completions and /chat/completions like:

```
openapi_extra={
        "x-fern-streaming": {
            "stream-condition": "$request.stream",
            "response": {
                "$ref": "#/components/schemas/OpenAICompletion"
            },
            "response-stream": {
                "$ref": "#/components/schemas/OpenAICompletion"
            }
        }
    }
```

openapi.json is generated automatically using `make api-docs`

Let me know if you are ok doing this change. Otherwise I can do it in a different PR

dsinghvi:
> Instead of modifying openapi.json directly, we'd need to add the extension to FastAPI endpoint definition in both /completions and /chat/completions like:
> 
> ```
> openapi_extra={
>         "x-fern-streaming": {
>             "stream-condition": "$request.stream",
>             "response": {
>                 "$ref": "#/components/schemas/OpenAICompletion"
>             },
>             "response-stream": {
>                 "$ref": "#/components/schemas/OpenAICompletion"
>             }
>         }
>     }
> ```
> 
> openapi.json is generated automatically using `make api-docs`
> 
> Let me know if you are ok doing this change. Otherwise I can do it in a different PR

Will do -- We should have an update tomorrow for you!

PR #1277, opened 2023-11-19
Add docker compose

evansnguyen:
Add docker compose to Private_GPT in docker env

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -29,8 +29,6 @@ RUN poetry install --with ui
 FROM base as app
 
 ENV PYTHONUNBUFFERED=1
-ENV PORT=8080
-EXPOSE 8080
 
 # Prepare a non-root user
 RUN adduser --system worker
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,11 +1,10 @@
 services:
   private-gpt:
     build:
-      context: ../privategpt
       dockerfile: Dockerfile.local
     volumes:
-      - ../privategpt/local_data/:/home/worker/app/local_data
-      - ../privategpt/models/:/home/worker/app/models
+      - ./local_data/:/home/worker/app/local_data
+      - ./models/:/home/worker/app/models
     ports:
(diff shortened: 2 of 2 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
+services:
+  private-gpt:
+    build:
+      dockerfile: Dockerfile.local
+    volumes:
+      - ./local_data/:/home/worker/app/local_data
+      - ./models/:/home/worker/app/models
+    ports:
+      - 127.0.0.1:8080:8001
```
pabloogc, line 9: We chose the default 8001 for end users, swapping it to 8080 will just confuse people

```suggestion
      - 127.0.0.1:8001:8001
```

docker-compose.yaml
```diff
@@ -0,0 +1,14 @@
...
+    environment:
+      PORT: 8001
```
pabloogc, line 11: Here you actually want to use the usual 8080

```suggestion
      PORT: 8080
```

pabloogc (changes requested):

docker-compose.yaml
evansnguyen: thanks for the explanation, I swapped the ports as you suggested

docker-compose.yaml
evansnguyen: done

gianniacquisto (approved):
This looks like a good addition

Dockerfile.local
```diff
@@ -30,7 +30,7 @@ FROM base as app
 
 ENV PYTHONUNBUFFERED=1
 ENV PORT=8080
-EXPOSE 8080
+EXPOSE 8001
```
pabloogc, line 33: The exposed port is still 8080, the docker compose is just mapping it
```suggestion
EXPOSE 8080
```

pabloogc (changes requested):

pabloogc (approved):

Issue #1226, opened 2023-11-13
Bulk Local Ingestion problem.


frenchiveruti:
>Bulk Local Ingestion
> 
> When you are running PrivateGPT in a fully local setup, you can ingest a complete folder for convenience (containing pdf, text files, etc.) and optionally watch changes on it with the command:
> 
> make ingest /path/to/folder -- --watch
> 
> To log the processed and failed files to an additional file, use:
> 
> make ingest /path/to/folder -- --watch --log-file /path/to/log/file.log
> 
> After ingestion is complete, you should be able to chat with your documents by navigating to http://localhost:8001/ and using the option Query documents, or using the completions / chat API.

Instead, I get:
```
llama_new_context_with_model: total VRAM used: 4857.93 MB (model: 4095.05 MB, context: 762.87 MB)
AVX = 1 | AVX2 = 1 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 1 | NEON = 0 | ARM_FMA = 0 | F16C = 1 | FP16_VA = 0 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 1 | SSSE3 = 0 | VSX = 0 |
Traceback (most recent call last):
  File "C:\Users\Fran\privateGPT\scripts\ingest_folder.py", line 80, in <module>
    raise ValueError(f"Path {args.folder} does not exist")
ValueError: Path `arg=C:\\Users\\Fran\\privateGPT\\FilesFeed does not exist
make: *** [Makefile:52: ingest] Error 1
```

No matter how I phrase the path. 
And I think it's because the line is being converted to "\`arg=(...)" that means, it's adding the `arg= ` section without reason.

rar8022:
So it's a "hack job" but it works if it helps you to solve it. I went into the ingest_folder.py and edited line 78: "path = Path(args.folder)" to "path = Path(args.folder[5:])" to strip out the arg= being added in. Then if I use the command make ingest "C:\Users\import" it works as intended. I've spent a few hours trying to figure out where the arg= is inserted and I just can't.

FrangSierra:
The `arg=` param comes from the `Makefile`. However the problem that you are probably facing if you are a Windows user is that you need to set the Args during the call on the command line.
Instead of 
`make ingest C:\Users\yourUsers\Documents\bulkFolder\`
 do 
`make ingest args =C:\Users\yourUsers\Documents\bulkFolder\`
 
[...]

PR #1289, opened 2023-11-21
Update ingest api docs with Windows support

FrangSierra:
Simple PR adding a note for parametrized `make` calls on Windows. 
Saw that some users were having some troubles making use of it on Windows command line here: #1226 (probably u can close this issue!)

If you prefer a different format to include this note on the docs just let me know and will modify it.
Thanks


FrangSierra:
I see that the preview-docs test from Fern is failing. However the trace doesnt give a lot of info about what's the problem. Is there any specific format for the documentation that I may have missed and im not applying?

lopagela (approved):
LGTM - modulo the preview error that Ivan is working on

Issue #1294, opened 2023-11-22
Local embeddings with OpenAI LLMs


Akbarable:
Hi everyone, 

I want to create a new PGPT profile that uses the local  `embedding_hf_model_name: BAAI/bge-small-en-v1.5` embedding model and `openai` LLM (e.g gpt-3.5 turbo etc.).

Is there a way I can do that? 

Looking for advice, thanks!

jochalek:
I got this working and made a PR #1296

PR #1279, opened 2023-11-19
Speed up document ingestion

lopagela:
* Refactor ingestion service to keep the index reference in memory

WIP

PR #1296, opened 2023-11-22
Support local embedding with OpenAI chat

jochalek:
Users can now locally embed documents and still choose to call out to OpenAI for chat completions.

Simply launching with the following settings.yaml

```yaml
llm:
  mode: hybrid
```

I understand using OpenAI is counter to the private aspect of privateGPT, but since OpenAI is supported already I found this useful for myself.

Also solves #1294

### DIFF
--- a/private_gpt/components/embedding/embedding_component.py
+++ b/private_gpt/components/embedding/embedding_component.py
@@ -21,7 +21,6 @@ class EmbeddingComponent:
                     cache_folder=str(models_cache_path),
                 )
             case "sagemaker":
-
                 from private_gpt.components.embedding.custom.sagemaker import (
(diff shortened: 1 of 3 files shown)


seanmavley:
I like this. Thanks for the PR. I'm gonna give it a try too

Considering the models are from HF, I was hoping support for HF inference would be included by default, rather aws sagemaker. Ha.

lopagela:
I've done something similar in #1279 -- I am splitting the configuration from the existing `llm.mode`  to `embedding.mode`, so that you can run whatever combination you want.

It will also come with performance improvements for the ingestion

jochalek:
I'm closing this in favor of #1279

PR #1301, opened 2023-11-23
Add multiple prompt style

lopagela:
Vigogne model for example does not work with llama2 model. Using the default `role: message` format from llama_index give good results.

### DIFF
--- a/private_gpt/settings/settings.py
+++ b/private_gpt/settings/settings.py
@@ -91,7 +91,26 @@ class VectorstoreSettings(BaseModel):
 class LocalSettings(BaseModel):
     llm_hf_repo_id: str
     llm_hf_model_file: str
-    embedding_hf_model_name: str
+    embedding_hf_model_name: str = Field(
+        description="Name of the HuggingFace model to use for embeddings"
+    )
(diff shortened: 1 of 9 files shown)


private_gpt/settings/settings.py
```diff
@@ -91,7 +91,26 @@ class VectorstoreSettings(BaseModel):
...
+    embedding_hf_model_name: str = Field(
+        description="Name of the HuggingFace model to use for embeddings"
+    )
+    prompt_style: Literal["default", "llama2", "tag"] = Field(
+        "llama2",
+        description=(
+            "The prompt style to use for the chat engine. "
+            "If `default` - use the default prompt style from the llama_index. It should look like `role: message`.\n"
+            "If `llama2` - use the llama2 prompt style from the llama_index. Based on `<s>`, `[INST]` and `<<SYS>>`.\n"
+            "If `tag` - use the tag prompt style from the llama_index. It should look like <|role|>: message. \n"
```
imartinez, line 103: This one is not coming from llama_index, right? So the description is not exact.

imartinez (commented):
Love it!!

imartinez:
Merged in https://github.com/imartinez/privateGPT/pull/1307

PR #1308, opened 2023-11-24
Chunk size

LuizFonseca06:


lopagela (changes requested):
There are no room for configuration, a regression regarding the way the vector store is created if it does not exist, and the new behavior is inconsistent depending on the present or not of an existing index.

Proposing to close this PR.

PR #1317, opened 2023-11-26
Added max_new_tokens as a config option to llm yaml block

gianniacquisto:
Added max_new_tokens as a configuration option to the llm block in settings

### DIFF
--- a/fern/docs/pages/manual/settings.mdx
+++ b/fern/docs/pages/manual/settings.mdx
@@ -83,8 +83,8 @@ server:
 
 The `llm` section of the settings allows for the following configurations:
 
-- mode = how to run your llm
-- max_new_tokens = this lets you configure the number of new tokens the llm will generate and add to the context window (by default Llama.cpp uses 256)
+- `mode`: how to run your llm
+- `max_new_tokens`: this lets you configure the number of new tokens the LLM will generate and add to the context window (by default Llama.cpp uses `256`)
 
 Example:
 
--- a/fern/docs/pages/manual/settings.mdx
+++ b/fern/docs/pages/manual/settings.mdx
@@ -78,3 +78,18 @@ Missing variables with no default will produce an error.
 server:
   port: ${PORT:8001}
 ```
+
+## LLM config options
+
+The `llm` section of the settings allows for the following configurations:
+
+- mode = how to run your llm
+- max_new_tokens = this lets you configure the number of new tokens the llm will generate and add to the context window (by default Llama.cpp uses 256)
+
+Example:
+
+```yaml
+llm:
+  mode: local
+  max_new_tokens: 256
+```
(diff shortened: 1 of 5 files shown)


fern/docs/pages/manual/settings.mdx
```diff
@@ -77,4 +77,19 @@ Missing variables with no default will produce an error.
...
 server:
   port: ${PORT:8001}
+```
+
+## LLM config options
+
+The `llm` section of the settings allows for the following configurations:
+
+- mode = how to run your llm
+- max_new_tokens = this lets you configure the number of new tokens the llm will generate and add to the context window (by default Llama.cpp uses 256)
```
lopagela, line 87: Using more markdown syntax to highlight the actual configuration key
```suggestion
- `mode`: how to run your llm
- `max_new_tokens`: this lets you configure the number of new tokens the LLM will generate and add to the context window (by default Llama.cpp uses `256`)
```

private_gpt/settings/settings.py
```diff
@@ -82,6 +82,7 @@ class DataSettings(BaseModel):
 
 class LLMSettings(BaseModel):
     mode: Literal["local", "openai", "sagemaker", "mock"]
+    max_new_tokens: int
```
lopagela, line 85: I believe the default should be defined in this file, instead of `settings.yaml`.
Could you also add the description of this parameter next to the default? C.f. below

```suggestion
    max_new_tokens: int = Field(256, description="The maximum number of token that the LLM is authorized to generate in one completion.")
```

lopagela (changes requested):
The default value should be defined in the code. `settings.yaml` should not be modified.

Putting default values in `settings.yaml` is making the file `settings.yaml` consistantly growing, and will make the creation of custom configuration harder for users - which is why we should try to keep the definition of default values from there.

fern/docs/pages/manual/settings.mdx
```diff
@@ -77,4 +77,19 @@ Missing variables with no default will produce an error.
 ```yaml
 server:
   port: ${PORT:8001}
+```
+
+## LLM config options
```
imartinez, line 82: I think this fits better in 

"Customizing low level parameters" within Installation

or in 

"Advanced setup" > "LLM Backends"

imartinez (changes requested):

private_gpt/settings/settings.py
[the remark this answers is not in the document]
gianniacquisto: Addressed this comment

fern/docs/pages/manual/settings.mdx
gianniacquisto: Good idea - i've moved it to the former section

imartinez (approved):

lopagela (approved):

PR #1318, opened 2023-11-26
Allow passing a system prompt

imartinez:
Allow passing a `system_prompt` in completions API and to send a role:system message in chat/completions APIs messages array to modify the LLM behaviour. 

- Added a basic `system_prompt` to Gradio Query Docs to avoid the LLM answering questions with context that is not in the documents.

### DIFF
--- a/private_gpt/server/chat/chat_service.py
+++ b/private_gpt/server/chat/chat_service.py
@@ -1,11 +1,10 @@
 from injector import inject, singleton
 from llama_index import ServiceContext, StorageContext, VectorStoreIndex
-from llama_index.chat_engine import ContextChatEngine
+from llama_index.chat_engine import ContextChatEngine, SimpleChatEngine
 from llama_index.chat_engine.types import (
     BaseChatEngine,
 )
 from llama_index.indices.postprocessor import MetadataReplacementPostProcessor
-from llama_index.llm_predictor.utils import stream_chat_response_to_tokens
 from llama_index.llms import ChatMessage
 from llama_index.types import TokenGen
 from pydantic import BaseModel
@@ -58,18 +57,23 @@ class ChatService:
         )
 
     def _chat_engine(
-        self, context_filter: ContextFilter | None = None
+        self, use_context: bool = False, context_filter: ContextFilter | None = None
     ) -> BaseChatEngine:
-        vector_index_retriever = self.vector_store_component.get_retriever(
-            index=self.index, context_filter=context_filter
-        )
-        return ContextChatEngine.from_defaults(
-            retriever=vector_index_retriever,
-            service_context=self.service_context,
-            node_postprocessors=[
-                MetadataReplacementPostProcessor(target_metadata_key="window"),
-            ],
-        )
+        if use_context:
+            vector_index_retriever = self.vector_store_component.get_retriever(
+                index=self.index, context_filter=context_filter
+            )
+            return ContextChatEngine.from_defaults(
+                retriever=vector_index_retriever,
+                service_context=self.service_context,
+                node_postprocessors=[
+                    MetadataReplacementPostProcessor(target_metadata_key="window"),
+                ],
+            )
+        else:
+            return SimpleChatEngine.from_defaults(
+                service_context=self.service_context,
+            )
 
     def stream_chat(
         self,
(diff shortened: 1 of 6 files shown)


imartinez:
@3ly-13 yes, I think it is a great idea to add it as additional inputs. Not only for the LLM chat but also for the Query Docs (in that case the input text should be initialized with the system prompt that is being used by default in Query Docs)

private_gpt/server/completions/completions_router.py
```diff
@@ -67,6 +72,7 @@ def prompt_completion(
     message = OpenAIMessage(content=body.prompt, role="user")
     chat_body = ChatBody(
         messages=[message],
+        system_prompt=body.system_prompt,
```
louis-jaris, line 75: Why not simply adding a message with the role `system` if a system prompt is passed? This remove the need for this extra parameter

private_gpt/server/chat/chat_service.py
```diff
@@ -58,63 +57,69 @@ def __init__(
...
+        )
         return completion_gen
 
     def chat(
         self,
         messages: list[ChatMessage],
+        system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
     ) -> Completion:
```
louis-jaris, line 112: `system_prompt` can be passed in `ChatMessage`, as the first message.

Doing this would reduce the number of moving pieces and the number of parameters in function.

This also contains the "state" exclusively in the list of messages

private_gpt/server/chat/chat_service.py
```diff
@@ -58,63 +57,69 @@ def __init__(
...
+            vector_index_retriever = self.vector_store_component.get_retriever(
+                index=self.index, context_filter=context_filter
             )
-            sources = [
-                Chunk.from_node(node) for node in streaming_response.source_nodes
-            ]
-            completion_gen = CompletionGen(
-                response=streaming_response.response_gen, sources=sources
+            return ContextChatEngine.from_defaults(
+                system_prompt=system_prompt,
```
louis-jaris, line 70: `system_prompt` here is really a `default_system_prompt`

louis-jaris (commented):
While I understand the will to be able to set a default system prompt, however it should be named accordingly

Additionally, in the new OpenAI API (c.f. https://platform.openai.com/docs/api-reference/assistants/modifyAssistant), this concept seems to be named `instructions`

----

In order to keep implementing the OpenAI API, I believe we should be using the parameter `instructions` instead of `system_prompt`.
To be confirmed, but we should also document that this prompt will not be taken into account if the first message in a `system` message.

private_gpt/server/completions/completions_router.py
lopagela, line 75: Why not simply adding a message with the role `system` if a system prompt is passed? This remove the need for this extra parameter

private_gpt/server/chat/chat_service.py
```diff
@@ -58,63 +57,69 @@ def __init__(
...
+        sources = [Chunk.from_node(node) for node in streaming_response.source_nodes]
+        completion_gen = CompletionGen(
+            response=streaming_response.response_gen, sources=sources
+        )
         return completion_gen
 
     def chat(
         self,
         messages: list[ChatMessage],
+        system_prompt: str | None = None,
```
lopagela, line 109: `system_prompt` can be passed in `ChatMessage`, as the first message.

Doing this would reduce the number of moving pieces and the number of parameters in function.

This also contains the "state" exclusively in the list of messages

private_gpt/server/chat/chat_service.py
lopagela, line 70: `system_prompt` here is really a `default_system_prompt`

lopagela (commented):
While I understand the will to be able to set a default system prompt, however it should be named accordingly

Additionally, in the new OpenAI API (c.f. https://platform.openai.com/docs/api-reference/assistants/modifyAssistant), this concept seems to be named `instructions`

----

In order to keep implementing the OpenAI API, I believe we should be using the parameter `instructions` instead of `system_prompt`.
To be confirmed, but we should also document that this prompt will not be taken into account if the first message in a `system` message.

private_gpt/server/chat/chat_service.py
imartinez: What do you mean? It is not a default system prompt, it is the one passed by the user.

imartinez:
@lopagela I've researched deeper the idea of using a system message in the messages array of the input instead of this new field. But the problem is that it won't work for contextualized completions (the ones using the context from documents) unless we stop using LlamaIndex's ContextChatEngine altogether and implement it ourselves -and I don't think that's a good idea at this point.

The way ChatEngine works in what has to do with system message is:
- Get context from docs
- Add the context in the SYSTEM block
- Add, after the context, the user sent system prompt within the SYSTEM block.

So essentially  the ChatEngine manages putting together the context and the user system_prompt. If you send it as a system message in the messages array, it fails because the ChatEngine doesn't know how to mix the context and the system prompt together properly.

imartinez:
@lopagela while `instructions`is being used in the new assistant API by OpenAPI, I feel system_prompt is better suited for the completions APIs as it is more explicit on the usage. I'd stick to it if you are ok with that.

3ly-13:
> @3ly-13 yes, I think it is a great idea to add it as additional inputs. Not only for the LLM chat but also for the Query Docs (in that case the input text should be initialized with the system prompt that is being used by default in Query Docs)

Great. I created #1329 to track this feature. It will likely be based on the changes in this PR once it's merged. Thanks!

lopagela (approved):

private_gpt/server/chat/chat_service.py
```diff
@@ -30,6 +31,40 @@ class CompletionGen(BaseModel):
     sources: list[Chunk] | None = None
 
 
+@dataclass
+class ChatEngineInput:
```
pabloogc, line 35: 👍

pabloogc (approved):

PR #1328, opened 2023-11-27
Zephyr prompt style

lopagela:
Coming from the example: https://github.com/run-llama/llama-hub/blob/7e5e3b1f74015ab0fc10eae2f903362bf57f3946/llama_hub/llama_packs/zephyr_query_engine/base.py#L37

Please note that you **may have to** remove to system prompt to fit to the example above


gianniacquisto:
Can you explain how the prompt styles are used? :)

PR #1336, opened 2023-11-30
Fix the parallel ingestion mode, and make it available through conf

lopagela:
Also updated the documentation to show how to configure the ingest mode.

### DIFF
--- a/settings.yaml
+++ b/settings.yaml
@@ -25,6 +25,9 @@ llm:
 embedding:
   # Should be matching the value above in most cases
   mode: local
+  ingest_mode: simple
+#  ingest_mode: parallel
+#  count_workers: 2
 
 vectorstore:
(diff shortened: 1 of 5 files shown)


settings.yaml
```diff
@@ -25,6 +25,9 @@ llm:
 embedding:
   # Should be matching the value above in most cases
   mode: local
+  ingest_mode: simple
+#  ingest_mode: parallel
```
imartinez, line 29: I'd try not to have commented out entries in settings.yaml. Maybe a comment pointing to the documentation section would be useful though.

imartinez (dismissed):

settings.yaml
lopagela: Done

imartinez (approved):

Issue #1329, opened 2023-11-27
Support user provided system prompts


3ly-13:
Add support for users to enter custom system prompts for the all chat modes. This could be added as an "Additional Options" field and adds a lot of flexibility in tailoring the system for specific use cases (see attached proof of concept). 

This can be implemented after #1318 is closed.

<img width="782" alt="C++-Prompt" src="https://github.com/imartinez/privateGPT/assets/143585971/dd0f020f-378e-48c9-af4c-b5149cb95090">
<img width="766" alt="Python-Prompt" src="https://github.com/imartinez/privateGPT/assets/143585971/8fe21319-41fd-4d85-905a-fc7dbd8bd73f">

3ly-13:
This is now complete with #1353. This issue can be closed.

PR #1353, opened 2023-12-01
Allows User to Set System Prompt via "Additional Options" in Chat Interface

3ly-13:
This PR allows users to set the system prompt from within the gradio interface (#1329). The default system prompt shows up as a `placeholder` to the input box, and the user may choose to override it.

I plan to add some documentation updates that provide example prompts, and these updates could be linked to from the gradio interface.

https://github.com/imartinez/privateGPT/assets/143585971/8f8d9580-f93e-4f90-be71-c08b2fbe2db2


fern/docs/pages/manual/ui.mdx
```diff
@@ -35,5 +35,30 @@ database* section in the documentation.
...
+
+##### System Prompt Examples:
+
+The system prompt can effectively provide your chat bot specialized roles, and results tailored to the prompt
+you have given the model. Examples of system prompts can be be found
+[here](https://www.w3schools.com/gen_ai/chatgpt-3-5/chatgpt-3-5_roles.php).
+
+Some interesting examples to try include:
+
+* You are <X>. You have all the knowledge and personality of <X>. Answer as if you were Shakespeare using
```
imartinez, line 55: I understand you don't want to say "Shakespear" here but a generic <X> to make the example fully generic.

private_gpt/ui/ui.py
```diff
@@ -114,25 +120,22 @@ def build_history() -> list[ChatMessage]:
 
         new_message = ChatMessage(content=message, role=MessageRole.USER)
         all_messages = [*build_history(), new_message]
+        # Add a system message to force the behaviour of the LLM
```
imartinez, line 123: Given the system prompt can be about anything, this specific comment does not apply anymore

imartinez (changes requested):

private_gpt/ui/ui.py
```diff
@@ -154,6 +157,32 @@ def build_history() -> list[ChatMessage]:
...
                 )
 
+    # On initialization and on mode change, this function set the system prompt
+    # to the default prompt based on the mode (and user settings).
+    @staticmethod
+    def _get_default_system_prompt(mode: str) -> str:
+        p = ""
+        match mode:
+            case "Query Docs":
+                p = "You can only answer questions about the provided context. If you know the answer " \
```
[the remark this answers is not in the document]
3ly-13: Thanks... The challenge is you may want 2 separate settings fields (1 for default_query_system_prompt, and 1 for default_chat_system_prompt), since the default prompts for those vary. 

I can go ahead and make those 2 default prompts explicitly defined in the settings, but the query prompt has no other default defined (unlike chat prompt, which can use `llama_index.DEFAULT_SYSTEM_PROMPT`).

Do you recommend we provide the default query prompt in the settings.yaml, and have it as a required field, or should we hard code a default value in the code, and allow users to override via settings?

fern/docs/pages/manual/ui.mdx
3ly-13: Fixed in latest commit.

private_gpt/ui/ui.py
3ly-13: Updated comment to `# If a system prompt is set, add it as a system message` and added code to not add the system message if no system prompt is defined.

private_gpt/ui/ui.py
```diff
@@ -154,6 +157,32 @@ def build_history() -> list[ChatMessage]:
...
+                    "but it is not based in the provided context, don't provide the answer, just state " \
+                    "the answer is not in the context provided."
+            case "LLM Chat":
+                p = settings().local.default_system_prompt or llama_utils.DEFAULT_SYSTEM_PROMPT
+            case "Search in Docs":
+                # TODO - Verify no prompt needed for doc search (and no default prompt options)
```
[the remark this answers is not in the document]
3ly-13: Updated to account for the general case of modes that do not support prompts. In that case the `system_prompt` variable will be cleared, and the UI will not allow entry of prompts for consistency.

private_gpt/ui/ui.py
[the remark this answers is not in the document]
3ly-13: I've pushed the solution above for testing (2 different settings fields) for evaluation. I'm open to another approach if there's a more efficient or clear solution. Thanks!

lopagela (approved):
Looks good to me !

imartinez (approved):

fern/docs/pages/manual/ui.mdx
```diff
@@ -35,5 +35,33 @@ database* section in the documentation.
...
 Normal chat interface, self-explanatory ;)
 
-You can check the actual prompt being passed to the LLM by looking at the logs of
-the server. We'll add better observability in future releases.
\ No newline at end of file
+#### System Prompt
+You can view and change the system prompt being passed to the LLM by clicking "Additional Inputs"
+in the chat interface. The system prompt is also logged on the server.
+
+By default, the `Query Docs` mode uses the setting value `local.default_query_system_prompt`.
```
imartinez, line 42: Better remove this as we are going to remove default_query_system_prompt

fern/docs/pages/manual/ui.mdx
```diff
@@ -35,5 +35,33 @@ database* section in the documentation.
...
+
+The `LLM Chat` mode attempts to use the optional settings value `local.default_chat_system_prompt`,
```
imartinez, line 44: Better remove this as we are going to remove default_chat_system_prompt

private_gpt/settings/settings.py
```diff
@@ -108,6 +108,19 @@ class LocalSettings(BaseModel):
             "`llama2` is the historic behaviour. `default` might work better with your custom models."
         ),
     )
+    default_chat_system_prompt: str | None = Field(
+        None,
+        description=(
+            "The default system prompt to use for the chat mode. "
+            "If none is given - use the default system prompt (from the llama_index). "
```
imartinez, line 115: Let's not fall back to llama_index DEFAULT_SYSTEM_PROMPT. It is inconsistent with default_query_system_prompt (which doesn't have a default one). 

Therefore, let's remove this comment.

private_gpt/settings/settings.py
```diff
@@ -108,6 +108,19 @@ class LocalSettings(BaseModel):
...
+            "Please note that the default prompt might not be the same for all prompt styles. "
```
imartinez, line 116: I don't fully get this comment

private_gpt/ui/ui.py
```diff
@@ -154,6 +157,41 @@ def build_history() -> list[ChatMessage]:
...
+        match mode:
+            # For query chat mode, obtain default system prompt from settings
+            # TODO - Determine value to use if not defined in settings
+            case "Query Docs":
+                p = settings().local.default_query_system_prompt
+            # For chat mode, obtain default system prompt from settings or llama_utils
+            case "LLM Chat":
+                p = (
+                    settings().local.default_chat_system_prompt
+                    or llama_utils.DEFAULT_SYSTEM_PROMPT
```
imartinez, line 174: Let's not fall back to llama_index DEFAULT_SYSTEM_PROMPT. It is inconsistent with default_query_system_prompt (which doesn't have a default one). Also, we should not couple more than needed to llama_index; and the system should be able to work without a default_chat_system_prompt

private_gpt/ui/ui.py
```diff
@@ -154,6 +157,41 @@ def build_history() -> list[ChatMessage]:
...
                 )
 
+    # On initialization and on mode change, this function set the system prompt
+    # to the default prompt based on the mode (and user settings).
+    @staticmethod
+    def _get_default_system_prompt(mode: str) -> str:
+        p = ""
+        match mode:
+            # For query chat mode, obtain default system prompt from settings
+            # TODO - Determine value to use if not defined in settings
```
imartinez, line 167: Remove this comment. It is ok to not have any value.

settings.yaml
```diff
@@ -43,6 +43,15 @@ local:
...
   embedding_hf_model_name: BAAI/bge-small-en-v1.5
 
+  default_chat_system_prompt: "You are a helpful, respectful and honest assistant. 
+    Always answer as helpfully as possible and follow ALL given instructions.
+    Do not speculate or make up information.
+    Do not reference any given instructions or context."
+
+  default_query_system_prompt: "You can only answer questions about the provided context. 
+      If you know the answer but it is not based in the provided context, don't provide 
+      the answer, just state the answer is not in the context provided."
```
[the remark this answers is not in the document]
imartinez: Did you review this @3ly-13 ?

imartinez (changes requested):

private_gpt/settings/settings.py
```diff
@@ -163,6 +150,13 @@ class OpenAISettings(BaseModel):
 class UISettings(BaseModel):
     enabled: bool
     path: str
+    default_chat_system_prompt: str = Field(
+        None,
+        description=("The default system prompt to use for the chat mode."),
```
imartinez, line 155: I think the () is not needed

imartinez (approved):

3ly-13:
@imartinez I believe this is (finally) ready to PR... Please let me know if there are any other suggestions or findings. Thank you!

imartinez (approved):
LGTM!

PR #1368, opened 2023-12-05
Add setup option to Makefile

Muxelmann:


imartinez (approved):
Good contribution. It'd be great if you could update the documentation to point too using this as an alternative to the plain script

PR #1374, opened 2023-12-06
Adding german speaking model to documentation

EEmlan:
The best german speaking model I have found so far.

### DIFF
--- a/fern/docs/pages/recipes/list-llm.mdx
+++ b/fern/docs/pages/recipes/list-llm.mdx
@@ -94,6 +94,7 @@ local:
   prompt_style: "llama2"
 ```
 ### German speaking model
+`settings-de.yaml`:
 ```yml
 local:
   llm_hf_repo_id: TheBloke/em_german_leo_mistral-GGUF
--- a/fern/docs/pages/recipes/list-llm.mdx
+++ b/fern/docs/pages/recipes/list-llm.mdx
@@ -93,7 +93,7 @@ local:
   embedding_hf_model_name: BAAI/bge-large-en
   prompt_style: "llama2"
 ```
-### Good German speaking Models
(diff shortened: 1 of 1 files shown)


Muxelmann:
I also played around with the following models:

- [TheBloke/openinstruct-mistral-7B-GGUF](https://huggingface.co/TheBloke/openinstruct-mistral-7B-GGUF)
- [TheBloke/claude2-alpaca-13B-GGUF](https://huggingface.co/TheBloke/claude2-alpaca-13B-GGUF)
- [TheBloke/neural-chat-7B-v3-2-GGUF](https://huggingface.co/TheBloke/neural-chat-7B-v3-2-GGUF)
- [TheBloke/deepseek-llm-7B-chat-GGUF](https://huggingface.co/TheBloke/deepseek-llm-7B-chat-GGUF)

If someone has a proper AI workstation with more graphics memory, maybe these are worth trying, too?
- [TheBloke/guanaco-33B-GGUF](https://huggingface.co/TheBloke/guanaco-33B-GGUF)
- [TheBloke/tigerbot-70B-chat-v2-GGUF](https://huggingface.co/TheBloke/tigerbot-70B-chat-v2-GGUF)

To that end, I wonder if we can/should add more prompt styles? The current options of `llama2`, `default` and `tag` as explained [here](https://docs.privategpt.dev/recipes/choice-of-llm/list-of-ll-ms) seem not to cover the above models - or I am setting it up incorrectly. That said, I am always using `BAAI/bge-small-en-v1.5` as an embedding model, maybe that's wrong, too.

imartinez:
We could definitely add more prompt styles. The current way of managing prompt styles is extensible. If you have any style in mind, please share it or open a PR with the contribution. Thanks!!

PR #1385, opened 2023-12-08
feat(llm): drop default_system_prompt

lopagela:
As discussed on Discord, the decision has been made to remove the system prompts by default, to better segregate the API and the UI usages.

A concurrent PR (#1353) is enabling the dynamic setting of a system prompt in the UI.

Therefore, if UI users want to use a custom system prompt, they can specify one directly in the UI. If the API users want to use a custom prompt, they can pass it directly into their messages that they are passing to the API.

In the highlight of the two use case above, it becomes clear that default system_prompt does not need to exist.

### DIFF
--- a/private_gpt/components/llm/llm_component.py
+++ b/private_gpt/components/llm/llm_component.py
@@ -23,10 +23,7 @@ class LLMComponent:
             case "local":
                 from llama_index.llms import LlamaCPP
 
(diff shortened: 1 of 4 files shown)


3ly-13:
I reviewed this PR and it looks good. I merged this with #1353 locally to test, and everything was functioning. There were minor merge conflicts to resolve which I can help with if needed.

imartinez (approved):

PR #1391, opened 2023-12-10
using `make setup` instead of poetry

Muxelmann:


lopagela (approved):

imartinez:
`make` is an optional dependency, so we should keep in the documentation the fallback to using `poetry run python scripts/setup`

imartinez (changes requested):
make is an optional dependency, so we should keep in the documentation the fallback to using poetry run python scripts/setup

imartinez:
Does not apply to the new docs. Please reopen adapting to the new docs. Thanks!

PR #1394, opened 2023-12-11
docs: remove misleading comment about pgpt working with python 3.12

eliottness:
I was misled into believing I could install using python 3.12 whereas the pyproject.toml explicitly states otherwise. This PR only removes this comment to make sure other people are not also trapped 😄


eliottness:
image
The preview-docs feature seems to still points to the version on `main`. Is this working properly ?

lopagela (approved):

imartinez (approved):

PR #1398, opened 2023-12-13
feat: Context Template from settings for QueryDocs Allowing the text …

juan-m12i:
For the QueryDocs use case, allow to customise the template used to collate the context pulled from the vector database.

- Added new config to Settings as a str? in rag
- Added context_template as parameter in chat_engine, stream_chat, and chat. If context_template is None, we load it from the new settings (and if there is no setting, it will get the default from llama_index as it was the previous behaviour)

[Github project link](https://github.com/users/imartinez/projects/3?pane=issue&itemId=44476632)
[Discord thread](https://discord.com/channels/1164200432894234644/1182746622413770822)

[though process](https://gist.github.com/juan-m12i/2b4fddeba1562bb98c17a72b79a933f0)

### DIFF
--- a/private_gpt/server/chat/chat_service.py
+++ b/private_gpt/server/chat/chat_service.py
@@ -19,6 +19,9 @@ from private_gpt.components.vector_store.vector_store_component import (
 )
 from private_gpt.open_ai.extensions.context_filter import ContextFilter
 from private_gpt.server.chunks.chunks_service import Chunk
+from private_gpt.settings.settings import settings
+
+DEFAULT_CONTEXT_TEMPLATE = settings().rag.default_context_template
 
 
 class Completion(BaseModel):
@@ -97,6 +100,7 @@ class ChatService:
         system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = DEFAULT_CONTEXT_TEMPLATE,
     ) -> BaseChatEngine:
         if use_context:
             vector_index_retriever = self.vector_store_component.get_retriever(
@@ -109,6 +113,7 @@ class ChatService:
                 node_postprocessors=[
                     MetadataReplacementPostProcessor(target_metadata_key="window"),
                 ],
+                context_template=context_template,
             )
         else:
             return SimpleChatEngine.from_defaults(
@@ -121,6 +126,7 @@ class ChatService:
         messages: list[ChatMessage],
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
     ) -> CompletionGen:
         chat_engine_input = ChatEngineInput.from_messages(messages)
         last_message = (
@@ -141,6 +147,7 @@ class ChatService:
             system_prompt=system_prompt,
             use_context=use_context,
             context_filter=context_filter,
+            context_template=context_template,
         )
         streaming_response = chat_engine.stream_chat(
             message=last_message if last_message is not None else "",
@@ -157,6 +164,7 @@ class ChatService:
         messages: list[ChatMessage],
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
     ) -> Completion:
         chat_engine_input = ChatEngineInput.from_messages(messages)
         last_message = (
@@ -177,6 +185,7 @@ class ChatService:
             system_prompt=system_prompt,
             use_context=use_context,
             context_filter=context_filter,
+            context_template=context_template,
         )
         wrapped_response = chat_engine.chat(
             message=last_message if last_message is not None else "",
--- a/settings.yaml
+++ b/settings.yaml
@@ -57,3 +57,6 @@ sagemaker:
 openai:
   api_key: ${OPENAI_API_KEY:}
   model: gpt-3.5-turbo
+
+rag:
+  default_context_template: "Context information is below.\n--------------------\n{context_str}\n--------------------\n"
--- a/private_gpt/settings/settings.py
+++ b/private_gpt/settings/settings.py
@@ -217,6 +217,16 @@ class QdrantSettings(BaseModel):
     )
 
 
+class RagSettings(BaseModel):
+    default_context_template: str | None = Field(
+        None,
+        description=(
+            "The default context template to use for the chat engine when using RAG. "
+            "If none is given - use the default system prompt (from the llama_index). "
+        ),
+    )
+
+
 class Settings(BaseModel):
     server: ServerSettings
     data: DataSettings
@@ -228,6 +238,7 @@ class Settings(BaseModel):
     openai: OpenAISettings
     vectorstore: VectorstoreSettings
     qdrant: QdrantSettings | None = None
+    rag: RagSettings
 
 
 """


settings.yaml
```diff
@@ -56,4 +56,7 @@ sagemaker:
 
 openai:
   api_key: ${OPENAI_API_KEY:}
-  model: gpt-3.5-turbo
\ No newline at end of file
+  model: gpt-3.5-turbo
+
+rag:
+  default_context_template: "Context information is below.\n--------------------\n{context_str}\n--------------------\n"
```
lopagela, line 62: Please use `|` (c.f. https://github.com/imartinez/privateGPT/pull/1403)

private_gpt/server/chat/chat_service.py
```diff
@@ -97,6 +100,7 @@ def _chat_engine(
         system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = DEFAULT_CONTEXT_TEMPLATE,
```
lopagela, line 103: The default value should be injected in the body of the function instead of here (given that it is coming from `settings()`) to make testing easier.

private_gpt/server/chat/chat_service.py
lopagela, line 1: Were is set (at usage) this new `context_template`?

lopagela (commented):
This addition seems similar to the recent work we've done around `(default_)system_prompt` in the sense that this `context_template` is a static value that will be applied to every queries.

I'll let @imartinez decide what is the best way forward

settings.yaml
juan-m12i: Thanks for the comment. 

Happy to do so, just one thing. In this case I was trying to keep the value as closely as possible as the original one, that I understand includes "\n" (see https://github.com/run-llama/llama_index/blob/main/llama_index/chat_engine/context.py)

DEFAULT_CONTEXT_TEMPLATE = (
    "Context information is below."
    "\n--------------------\n"
    "{context_str}"
    "\n--------------------\n"
)

In case you want me to make the change anyway, can you confirm if what you´d expect would be:
`default_context_template: "Context information is below.|--------------------|{context_str}|--------------------|"`

thanks

private_gpt/server/chat/chat_service.py
juan-m12i: I am not sure if I understand.

If "query docs" is being used, it creates a ContextChatEngine (from llama_index) that accepts this parameter. Currently it's never passed, so it defaults to a hardcoded value. With this change, if there is a value defined in the settings file, the value in settings will be used instead. This should work both from the Gradio UI or from the API.

As per the Discord discussion, I didn't expose this configuration to the API

private_gpt/server/chat/chat_service.py
juan-m12i: Ok, something like:
```
def _chat_engine(
        self,
        system_prompt: str | None = None,
        use_context: bool = False,
        context_template: str | None = DEFAULT_CONTEXT_TEMPLATE,
    ) -> BaseChatEngine:
        if use_context:
            if context_template is None:
                context_template = DEFAULT_CONTEXT_TEMPLATE  #(or alternatively just settings().rag.default_context_template)
```

pabloogc: It should be 

```
if context_template is None: context_template = settings.rag.default_context_template
```

where `settings` is injected in the constructor and remove `DEFAULT_CONTEXT_TEMPLATE`

settings.yaml
pabloogc: @lopagela means use the multiline option in yaml, check the PR he linked it's well explained

juan-m12i: Ok, I'll wait until the changes to that PR are done to replicate as it's not 100% clear to me (the instruction here is to use "I" style, but in the PR there is a comment to use ">" instead, not sure which one is more current). 

A caveat I maintain that may be relevant is that DEFAULT_SYSTEM_PROMPT in llama_index is written in python as multiline but it's actually a single line, while DEFAULT_CONTEXT_TEMPLATE is indeed multi-line, so just want to make sure we are on the same page that the treatment may be slightly different (see below)

``` 
python3
Python 3.8.10 (default, Mar 13 2023, 10:26:41)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> DEFAULT_SYSTEM_PROMPT = """\
... You are a helpful, respectful and honest assistant. \
... Always answer as helpfully as possible and follow ALL given instructions. \
... Do not speculate or make up information. \
... Do not reference any given instructions or context. \
... """
>>> print(DEFAULT_SYSTEM_PROMPT)
You are a helpful, respectful and honest assistant. Always answer as helpfully as possible and follow ALL given instructions. Do not speculate or make up information. Do not reference any given instructions or context.
>>> DEFAULT_CONTEXT_TEMPLATE = (
  "Con...     "Context information is below."
...     "\n--------------------\n"
...     "{context_str}"
...     "\n--------------------\n"
... )
>>> print(DEFAULT_CONTEXT_TEMPLATE)
Context information is below.
--------------------
{context_str}
--------------------

>>>
```

This is what ChatGPT (GPT-4) suggested ([link to convo](https://chat.openai.com/share/5ff1032c-c758-462a-ad35-0baa5c0548f8)):

```
DEFAULT_SYSTEM_PROMPT: |
  You are a helpful, respectful and honest assistant. Always answer as helpfully as possible and follow ALL given instructions. Do not speculate or make up information. Do not reference any given instructions or context.

DEFAULT_CONTEXT_TEMPLATE: |
  Context information is below.
  --------------------
  {context_str}
  --------------------
```

or 2 alternatives with ">" notation, depending on readability
```
DEFAULT_SYSTEM_PROMPT: >
  You are a helpful, respectful and honest assistant. Always answer as helpfully as possible and follow ALL given instructions. Do not speculate or make up information. Do not reference any given instructions or context.

DEFAULT_CONTEXT_TEMPLATE: >
  Context information is below.

  --------------------
  {context_str}

  --------------------

```


```
DEFAULT_SYSTEM_PROMPT: >
  You are a helpful, respectful and honest assistant. \
  Always answer as helpfully as possible and follow ALL given instructions. \
  Do not speculate or make up information. \
  Do not reference any given instructions or context.

DEFAULT_CONTEXT_TEMPLATE: >
  Context information is below.

  --------------------

  {context_str}

  --------------------
```

imartinez: The PR has been merged. In that case we wanted single line (as default_system_prompt is supposed to be single line). In the case of default_context_template you do want to have line breaks, so you should use | instead.

imartinez (changes requested):

imartinez:
> This addition seems similar to the recent work we've done around `(default_)system_prompt` in the sense that this `context_template` is a static value that will be applied to every queries.
> 
> I'll let @imartinez decide what is the best way forward

I see this kind of customization as a refinement of the default RAG pipeline, so I proposed to create a new `rag` section in settings for this - in the future it could host other straightforward customizations like the size of the chunk generated during ingestion. 

My overall view is:
- Allow some **simple modifications** to the default RAG pipeline through this new set of **settings**.
- For more **complex customizations**, direct the users to **fork** the repo and modify the code of Services.

juan-m12i:
Hi! I applied the requested changes.

The only difference from what @pabloogc mentioned is that I had to do: 
`settings().rag.default_context_template` instead of `settings.rag.default_context_template`, not sure if I should have done it differently, but it wouldn't work without parenthesis (and mypy would also complain)

private_gpt/server/chat/chat_service.py
```diff
@@ -97,8 +98,11 @@ def _chat_engine(
         system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
     ) -> BaseChatEngine:
         if use_context:
+            if context_template is None:
+                context_template = settings().rag.default_context_template
```
imartinez, line 105: You should avoid calling settings() directly, and instead inject settings as a dependency in the ChatService constructor. Check how settings are injected in EmbeddingComponent or LLMComponent for example.

private_gpt/server/chat/chat_service.py
```diff
@@ -97,8 +98,11 @@ def _chat_engine(
         system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
```
imartinez, line 101: Given we are not making use of this context_template argument anywhere, I'd suggest to remove it for the moment.

private_gpt/server/chat/chat_service.py
```diff
@@ -121,6 +126,7 @@ def stream_chat(
         messages: list[ChatMessage],
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
```
imartinez, line 129: Given we are not making use of this context_template argument anywhere, I'd suggest to remove it for the moment.

private_gpt/server/chat/chat_service.py
```diff
@@ -141,6 +147,7 @@ def stream_chat(
             system_prompt=system_prompt,
             use_context=use_context,
             context_filter=context_filter,
+            context_template=context_template,
```
imartinez, line 150: Given we are not making use of this context_template argument anywhere, I'd suggest to remove it for the moment.

private_gpt/server/chat/chat_service.py
```diff
@@ -97,8 +98,11 @@ def _chat_engine(
         system_prompt: str | None = None,
         use_context: bool = False,
         context_filter: ContextFilter | None = None,
+        context_template: str | None = None,
     ) -> BaseChatEngine:
         if use_context:
+            if context_template is None:
```
imartinez, line 104: Instead of this, why not checking if settings.rag.default_cotnext_template is not None and then apply it?

imartinez (changes requested):

juan-m12i:
Hi, replicated the settings approach from LLMcomponent and Embedding component (there were some differences as the setting is used in a private function of the class, but I believe the approach I took (creating an internal variable) should work.

I removed context_template from the params and function callings because as requested (I guess following YAGNI). My thinking process was to leave this prepared in such a way that if someone would use the class (ui, api, any other entrypoint) it could pass the parameter and keep a consistent interface, but to enable that back is such a small change that it's all good).

Only drawback is that by removing that, the conditional is now a bit more verbose, as the default None is not assigned:
```
            if self.default_context_template is not None:
                context_template = self.default_context_template
            else:
                context_template = None
```

Alternatively we could consider this one lines, but it's less readable IMO
`context_template = self.default_context_template if self.default_context_template is not None else None`

imartinez:
@juan-m12i could you update the branch to latest changes? I still think it is a valuable contribution!

jaluma (approved):

PR #1399, opened 2023-12-13
feat: Text to SQL support

ishaandatta:
Support for text-to-SQL for MySQL (https://github.com/users/imartinez/projects/3?pane=issue&itemId=44266590)

This uses `sqlalchemy` to initialise a client to the database.

Thereafter, a `SQLDatabase` instance is initialised, which is passed to the llama_index object `SQLTableNodeMapping` for creating a node per schema. (allowing query-time retrieval of schema)

Finally, the llama_index `SQLTableRetrieverQueryEngine` is used to-
    1. Convert text to appropriate SQL query
    2. Execute the SQL query
    3. Use the result as context to answer prompt


fern/docs.yml
```diff
@@ -58,6 +58,8 @@ navigation:
         contents:
           - page: Vector Stores
             path: ./docs/pages/manual/vectordb.mdx
+          - page: SQL Databases
+            path: ./docs/pages/manual/nlsql.mdx
```
lopagela, line 62: I don't think this should be under `Storage` section, but more into `Advance Setup` or something like this.

fern/docs/pages/manual/nlsql.mdx
```diff
@@ -0,0 +1,30 @@
...
+## Other Databases
+To get started with exploring the use of other databases, set the `sqldatabase.dialect` and `sqldatabase.driver` properties in the `settings.yaml` file.
+
+```yaml
+sqldatabase:
+  dialect: <dialect_here>
+  driver: <driver_here>
+```
+
+Refer to [SQLAlchemy Engine Configuration](https://docs.sqlalchemy.org/en/20/core/engines.html) for constructing the DB connection string.
```
lopagela, line 30: That would be great if that section could be a bit more user friendly - for example, putting additional links to SQLAlchemy reference (for example, the list of `dialect` https://docs.sqlalchemy.org/en/20/dialects/), and at least one additional link (to postgres or sqlite for example, which are very common).

private_gpt/components/nlsql/nlsql_component.py
```diff
@@ -0,0 +1,70 @@
...
+
+from injector import inject, singleton
+from llama_index import ServiceContext, SQLDatabase, VectorStoreIndex
+from llama_index.indices.struct_store import SQLTableRetrieverQueryEngine
+from llama_index.objects import ObjectIndex, SQLTableNodeMapping, SQLTableSchema
+from sqlalchemy import (
+    MetaData,
+)
+from sqlalchemy.engine import create_engine
+from sqlalchemy.engine.base import Engine
```
lopagela, line 12: Did you try without the `local` mode?

Are these dependencies shipped from `privateGPT` by default?

It looks to me that these additional dependencies are brought by default, while it should be an `opt-in` feature I think. This would also means that this `NLSQLComponent` should be conditionally injected in the app context (basically, the app should still be able to run without that SQL component).

Ideally, the app should load this component only if the configuration is specifying a valid SQL configuration (thanks to some "on/off" switch in configuration maybe)

private_gpt/settings/settings.py
```diff
@@ -217,6 +217,31 @@ class QdrantSettings(BaseModel):
     )
 
 
+class SQLDatabaseSettings(BaseModel):
+    enabled: bool = Field(
+        False,
+        description="Flag to enable SQL Query mode. Disabled by default",
+    )
+    dialect: Literal["mysql"]
+    driver: Literal["pymysql"]
```
lopagela, line 226: This is going against the documentation - Pydantic will fail to validate any other values, and consequently, one will not be able to change the dialect and driver.

private_gpt/settings/settings.py
```diff
@@ -217,6 +217,31 @@ class QdrantSettings(BaseModel):
...
+    db_host: str | None = Field(
+        "localhost",
+        description="Host name of Database server. If host is None, set to 'localhost'.",
+    )
+    db_user: str | None = Field(
+        "root",
+        description="Username to be used for accessing the SQL Database Server. If user is None, set to 'root'.",
+    )
```
lopagela, line 234: The default should be left to `None`, as `root` is not a standard default user in SQL norm (actually, SQL does not tell anything on a default user, but implementations are usually defining some, e.g. `postgres` for postgreSQL).

private_gpt/settings/settings.py
```diff
@@ -217,6 +217,31 @@ class QdrantSettings(BaseModel):
...
+    db_password: str = Field(
+        "",
+        description="Password to be used for accessing the SQL Database Server. If password is None, set to empty string.",
+    )
```
lopagela, line 238: The description is misleading. The default is a blank string, which is different from `None`

pyproject.toml
```diff
@@ -16,6 +16,9 @@ llama-index = { extras = ["local_models"], version = "0.9.3" }
 watchdog = "^3.0.0"
 qdrant-client = "^1.6.9"
 chromadb = {version = "^0.4.13", optional = true}
+gradio = "4.4.1"
+pymysql = "^1.1.0"
+cryptography = "^41.0.7"
```
lopagela, line 21: This has not its place in the default dependencies. `gradio` is in the `ui` target.

Additionally, I think that `pymsql` should not be added in the dependencies of this project, as this is a driver that is specific to mySQL: users might not use mySQL, and might need other dependencies (`psycopg2`, `psycopg`, and many others for example - c.f. sqlalchemy dialects)

settings.yaml
```diff
@@ -44,6 +44,16 @@ vectorstore:
...
 
+sqlmode:
+  enabled: false
+  dialect: mysql
+  driver: pymysql
+  db_host: ${HOSTNAME:localhost}
+  db_user: ${USERNAME:root}
+  db_password: ${PASSWORD:}
+  database: ${DATABASE:}
+  tables: ["${TABLES:}"]
```
lopagela, line 55: The env var are specific to SQL/DB world, so I'd personally prefix them with something like `SQL_` or `DB_`.

Additionally, `TABLES` is set to plural, while its usage shows that only a single name is supported, which is misleading.

lopagela (changes requested):
The current state of the Text to SQL support is not properly taking care of:
* Multi SQL dialect
* Execution **without Text to SQL** (i.e. Vanilla privateGPT) is broken
* Declaration of new dependencies is not justified

I think this feature must be tested across different "privateGPT mode" (with, and without DB support installed).

fern/docs/pages/manual/nlsql.mdx
```diff
@@ -0,0 +1,30 @@
...
+```yaml
+sqldatabase:
+  dialect: mysql
+  driver: pymysql
+  host: <db_hostname_here>
+  user: <db_username_here>
+  password: <db_password_here>
+  database: <database_name_here>
+  tables: <list_of_table_names_here>
+```
```
pabloogc, line 15: this is very rigid, why not move it to a parameter similar to `context_files` (like `context_database(s)`)? Conceptually speaking there is not big difference between talking to your documents or talking to your database(s)

I understand that would require a big refactor of this PR but it will set a pattern for next features (like `context_websites`)

settings.yaml
ishaandatta: The `tables` variable accepts a list of table names (you can setup multiple tables as context- `SQLTableRetrieverQueryEngine` will select the right table to query).

It's used in the code like- 
`for table_name in self.metadata_obj.tables:
            table_schema_objs.append(SQLTableSchema(table_name=table_name))`

lopagela: @ishaandatta you are right - `tables` does accept a list of string, but the env var `TABLES` seems to only accept one table.

Let me ask this: how would you pass several tables names in the env variable `TABLES`?

ishaandatta: Right, makes sense- I'm not actually sure how to define an array/list here in the env file.
Figured square brackets were enough as I tested and it works for multiple tables.

I tested it with the following value-
`tables: ["EMPLOYEE", "DEPARTMENT"]`

private_gpt/settings/settings.py
ishaandatta: MySQL default admin user is called `root`.

Regardless, wrt supporting other SQL databases, I'm making username and password both default to `None`

ishaandatta:
@lopagela addressed the changes, please check once

ricklettow:
Thank you for your efforts!  

I have hit a bump when running this on a large database schema.  It appears that SQLAlchemy is running a reflect() multiple times (1. during initial startup, 2. Before the first query in ObjectIndex.from_objects() ).  Subsequent queries perform better however the startup is > 30 minutes and the first query is another 30 minutes.  I am limiting to a single table however, that table has relationships to other tables that are recursively parsed.


```
sql_database = SQLDatabase(engine, include_tables=tables)

obj_index = ObjectIndex.from_objects(
          table_schema_objs,
          table_node_mapping,
          VectorStoreIndex,
          service_context=service_context,
      )
```

Both call above trigger a reflection.  Is this by design?  Can the reflection results be persisted for quick loading when the schema has infrequent changes?  It seems to be on the llama-index side but I was wondering if a different approach could be used.

Thank you.

AlfonsoLozana:

PR #1403, opened 2023-12-14
correct yaml multiline string

cognitivetech:
multiline string in yaml should not be surrounded in quote, instead `>` to condense string as single line or `|` to preserve formating.

https://yaml-multiline.info/

good for a first pull request :muscle:

### DIFF
--- a/settings.yaml
+++ b/settings.yaml
@@ -22,12 +22,12 @@ data:
 ui:
   enabled: true
   path: /
-  default_chat_system_prompt: |
+  default_chat_system_prompt: >
     You are a helpful, respectful and honest assistant. 
     Always answer as helpfully as possible and follow ALL given instructions.
     Do not speculate or make up information.
     Do not reference any given instructions or context.
-  default_query_system_prompt: |
+  default_query_system_prompt: >
     You can only answer questions about the provided context. 
     If you know the answer but it is not based in the provided context, don't provide 
     the answer, just state the answer is not in the context provided.
--- a/settings.yaml
+++ b/settings.yaml
@@ -22,13 +22,15 @@ data:
 ui:
   enabled: true
   path: /
-  default_chat_system_prompt: "You are a helpful, respectful and honest assistant. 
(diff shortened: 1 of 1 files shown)


settings.yaml
```diff
@@ -22,13 +22,15 @@ data:
...
+  default_chat_system_prompt: |
+    You are a helpful, respectful and honest assistant. 
     Always answer as helpfully as possible and follow ALL given instructions.
     Do not speculate or make up information.
-    Do not reference any given instructions or context."
-  default_query_system_prompt: "You can only answer questions about the provided context. 
-      If you know the answer but it is not based in the provided context, don't provide 
-      the answer, just state the answer is not in the context provided."
+    Do not reference any given instructions or context.
+  default_query_system_prompt: |
```
lopagela, line 30: Please use `>` over `|`

lopagela (commented):
Already raised https://github.com/imartinez/privateGPT/pull/1353#discussion_r1420879576

---

Could you please review the feedback and adjust? Thank you :pray:

cognitivetech:
> Already raised [#1353 (comment)](https://github.com/imartinez/privateGPT/pull/1353#discussion_r1420879576)
> 
> Could you please review the feedback and adjust? Thank you 🙏

to be clear, the comment https://github.com/imartinez/privateGPT/pull/1353#discussion_r1420879576 seems to imply that its desirable to preserve formatting (newlines), which is in line with the `|`. Rather than `>` which replaces newline with space.

I will change to `>` if that's what we want, just wish to be clear.

lopagela:
>  seems to imply that its desirable to preserve formatting (newlines),

I was maybe not that clear, that's right :sweat_smile: -- in this comment I meant that we do care to do not put new lines in the system prompt - so yes, `>` should be the way to go IMHO

cognitivetech:
ok, fine with me, code corrected. 

(maybe I should test this both ways.. even seems to do fine with the plain quotes we started with)

imartinez (approved):

PR #1415, opened 2023-12-17
Update default model to TheBloke/Mistral-7B-Instruct-v0.2-GGUF

imartinez:
- Make TheBloke/Mistral-7B-Instruct-v0.2-GGUF the default model in settings.yaml
- Update LlamaCPP dependency (it now works with modern models, including Mixtral, even though that one requires a 32GB RAM)
- Fix API empty documentation section


pabloogc (approved):

lopagela (approved):
:rocket:

PR #1417, opened 2023-12-17
Ingest plain text

imartinez:
Add a new API to ingest plain text. 
- Deprecate previous /ingest API
- Move previous ingestion logic to a new ingest/file API
- Create a new ingest/text API that allows for plain text ingestion by passing the file_name and the plain text to be ingested. A single Document is generated for the ingested text. 

All the previously existing ingestion logic is reused to leverage the different ingestion capabilities. A little refactor is done in IngestService to accommodate the new text ingestion

Screenshot 2023-12-17 at 22 13 08

### DIFF
--- a/fern/openapi/openapi.json
+++ b/fern/openapi/openapi.json
@@ -137,13 +137,98 @@
           "Ingestion"
         ],
         "summary": "Ingest",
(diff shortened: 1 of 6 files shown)


pabloogc (approved):
Looks good 👍

PR #1426, opened 2023-12-19
add mistral + chatml prompts

cognitivetech:
this adds what I believe is the correct prompt formatting for Mistral.. based on testing where I've rated it best vs llama-index and llama2

### DIFF
--- a/fern/docs/pages/recipes/list-llm.mdx
+++ b/fern/docs/pages/recipes/list-llm.mdx
@@ -24,7 +24,7 @@ user: {{ user_message }}
 assistant: {{ assistant_message }}
 ```
 
-And the "`tag`" style looks like this:
+The "`tag`" style looks like this:
 
 ```text
 <|system|>: {{ system_prompt }}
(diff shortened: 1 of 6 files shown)


louis-jaris (commented):
Tests are missing -- they will help maintainers to review the difference between `llama2` and `mistral` prompt format.

cognitivetech:
> Tests are missing -- they will help maintainers to review the difference between `llama2` and `mistral` prompt format.

i've added a test... which is failing, because it doesn't take into account my system prompt... which... 

I believe highlights the point that those other prompts aren't actually using the system prompt... just not sure what to do now.

cognitivetech:
I removed `def test_mistral_prompt_style_with_system_prompt():` because there aren't two expected behaviors. a given prompt method should _always_ add the system prompt.

However, I am not doing anything else until one of the more experienced maintainers can review my code \ thought process.

Maybe I'm missing something entirely

cognitivetech:
IDK what ruff wants of me.. I tired --fix and it didn't like the result, so I added space between the external vs internal imports, like in the original, still fail.

I'm a newb to python testing, but.. I will admit the test helped me fix some problems.

what with ruff, idk.

cognitivetech:
ok, all checks passed :sweat_smile:

cognitivetech:
ok, now I'm seeing the discussion where system prompt is only set for Gradio users https://github.com/imartinez/privateGPT/pull/1353

annnd, now I see in the api documentation where I can set system prompt as part of my api call :face_exhaling: 

okok... now I'm getting closer :sweat_smile: but .... at least I understand my confusion from before.

imartinez (approved):
Amazing 😍

Issue #1414, opened 2023-12-17
Docker compose file issues


NickDevoctomy:
Hi there,

I'm trying to get this running via docker-compose, unfortunately I don't see any explicit instructions anywhere so assume it's just a matter of running 'docker-compose up --build'.

Unfortunately after getting over the initial missing 'data' directory, I'm just chasing errors for missing files / folders. I assume there is something I'm supposed to have done prior to running docker compose?

Thanks for your time!

NickDevoctomy:
Okay, I found the solution on this page,

https://github.com/imartinez/privateGPT/issues/1251#issuecomment-1817994055

croossin:
@NickDevoctomy Were you able to push those models to a git repo? Trying to deploy this to portainer via a public git repo but having trouble since it exceeds the 2.00 GiB limit GitHub has.. trying to think of workarounds

NickDevoctomy:
I downloaded it manually and copied it into the folder.  From what I understand, it's supposed to have some kind of setup process that downloads the model automatically, but with the docker-compose, that doesn't seem to work.  Not sure why, but if that could be made to work that would sort out that issue.

lukaboljevic:
I was getting the same error, namely "Provided model path does not exist. Please check the path or provide a model_url to download" after running `docker compose up --build` with the default Dockerfile.local and docker compose file.

After implementing the changes in the above PR, the issue was solved and the app worked as expected, but it only runs on the CPU, which is also expected given the current content of the Dockerfile. I will try to get it to work on the GPU now.

NickDevoctomy:
Thanks @lukaboljevic I just tried your suggested changes out and it worked great, now utilising the GPU and automatic model download.

NickDevoctomy:
I had an issue with entrypoint.sh downloading the file each time regardless, so I modified the script slightly for a quick fix,

`
#!/bin/sh

# Check if the FILE_URL environment variable is set
if [ -z "$FILE_URL" ]
then
    echo "Error: FILE_URL environment variable is not set."
    exit 1
fi

# Extract filename from FILE_URL
FILENAME=$(basename "$FILE_URL")

# Define the path of the download and the complete file
DOWNLOAD_PATH="models/${FILENAME}"
COMPLETE_FILE="${DOWNLOAD_PATH}.complete"

# Check if the .complete file exists
if [ -f "$COMPLETE_FILE" ]
then
    echo "Download already completed for ${FILENAME}, skipping download."
else
[...]

PR #1428, opened 2023-12-19
Fix docker compose up and MacBook segfault

mbianchidev:
Implemented what was written on [this comment](https://github.com/imartinez/privateGPT/issues/1414#issuecomment-1859109980) and added some tweaks to make it work on without manual actions on the user's side

Could actually be a good idea to add a `Dockerfile.mac` instead of messing the `Dockerfile.local` file up, then it needs to be used in docker-compose.yaml when running on mac.

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -5,9 +5,9 @@ RUN pip install pipx
 RUN python3 -m pipx ensurepath
 RUN pipx install poetry
 ENV PATH="/root/.local/bin:$PATH"
-# Set the environment variable for the file URL (can be overridden)
+# Set the environment variable for the file URL (can be overwritten)
 ENV FILE_URL="https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf"
-# Set the predefined path where the file should be downloaded
+# Set the predefined model name (can be ovewritten)
 ENV NAME="mistral-7b-instruct-v0.1.Q4_K_M.gguf"
 
 # Dependencies to build llama-cpp
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,5 +1,3 @@
-### IMPORTANT, THIS IMAGE CAN ONLY BE RUN IN LINUX DOCKER
-### You will run into a segfault in mac
 FROM python:3.11.6-slim-bookworm as base
 
 # Install poetry
@@ -7,6 +5,10 @@ RUN pip install pipx
 RUN python3 -m pipx ensurepath
 RUN pipx install poetry
 ENV PATH="/root/.local/bin:$PATH"
+# Set the environment variable for the file URL (can be overridden)
+ENV FILE_URL="https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.1-GGUF/resolve/main/mistral-7b-instruct-v0.1.Q4_K_M.gguf"
+# Set the predefined path where the file should be downloaded
+ENV NAME="mistral-7b-instruct-v0.1.Q4_K_M.gguf"
 
 # Dependencies to build llama-cpp
 RUN apt update && apt install -y \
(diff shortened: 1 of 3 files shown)


imartinez:
The latest changes to dockerfiles merged address some of these issues, for example the download of models. Could you take a look and share your thoughts? https://github.com/imartinez/privateGPT/pull/1445

mbianchidev:
> The latest changes to dockerfiles merged address some of these issues, for example the download of models. Could you take a look and share your thoughts? #1445

@imartinez thanks a lot for checking in! 

Docker compose still doesn't work out-of-the-box cause the model is not downloaded by default, the only way to do it without embedding it in the image (ofc not feasible) is an entrypoint script, also there are two changes not present also propedeutic to make it work for the dockerfile around poetry.

`RUN poetry config installer.max-workers 10`
`RUN poetry install --extras chroma`

And for settings to switch the vectorstore to `chroma `instead of `qdrant`

```
vectorstore:
  database: chroma
```

Also `USER worker` before running the python command seems also to be unnecessary

The goal of my contribution is to make the project smooth for docker and docker compose users, if you clone the repo and build image + run or just run `docker compose up` one should be able to have it running locally with no further action.

Let me know if I can implement any changes to match more with your vision!

imartinez:
> > The latest changes to dockerfiles merged address some of these issues, for example the download of models. Could you take a look and share your thoughts? #1445
> 
> @imartinez thanks a lot for checking in!
> 
> Docker compose still doesn't work out-of-the-box cause the model is not downloaded by default, the only way to do it without embedding it in the image (ofc not feasible) is an entrypoint script, also there are two changes not present also propedeutic to make it work for the dockerfile around poetry.
> 
> `RUN poetry config installer.max-workers 10` `RUN poetry install --extras chroma`
> 
> And for settings to switch the vectorstore to `chroma` instead of `qdrant`
> 
> ```
> vectorstore:
>   database: chroma
> ```
> 
> Also `USER worker` before running the python command seems also to be unnecessary
> 
> The goal of my contribution is to make the project smooth for docker and docker compose users, if you clone the repo and build image + run or just run `docker compose up` one should be able to have it running locally with no further action.
> 
> Let me know if I can implement any changes to match more with your vision!

Thanks for the detailed explanation. 

- "model is not downloaded by default" -> there is already an entrypoint setup script that you can run after building the image that will take care of that. I think that's the best way to go:  docker compose run --rm --entrypoint="bash -c '[ -f scripts/setup ] && scripts/setup'" private-gpt
- poetry max worker is something we should add, I agree
- --extras chroma should not be added by default. Qdrant is the default vector db, and therefore Chroma-related dependencies are optional. Same for the change you propose to the settings.yaml file. 
- "USER worker" not sure about that one. I could test it out.

Let me know your thoughts.

mbianchidev:
> Thanks for the detailed explanation.
> 
> * "model is not downloaded by default" -> there is already an entrypoint setup script that you can run after building the image that will take care of that. I think that's the best way to go:  docker compose run --rm --entrypoint="bash -c '[ -f scripts/setup ] && scripts/setup'" private-gpt
> * poetry max worker is something we should add, I agree
> * --extras chroma should not be added by default. Qdrant is the default vector db, and therefore Chroma-related dependencies are optional. Same for the change you propose to the settings.yaml file.
> * "USER worker" not sure about that one. I could test it out.
> 
> Let me know your thoughts.

I think that the entrypoint as proposed is a bit verbose but this may be only my own preference.

I had to do a bit of research before getting up and running with compose and judging by some recent and similar issues opened I was not the only one encountering the errors.
What about adding the command: `docker compose run --rm --entrypoint="bash -c '[ -f scripts/setup ] && scripts/setup'" private-gpt` to both the readme file of the repo and the doc? 

A middle ground could also be creating a Dockerfile.mac to avoid touching the original one at all, what do you think?

Chroma is the only one working on mac/local if I'm not mistaken, correct? If that's the case I do think it might be better to enable it by default, what would the consequences be? Ofc it's a trade-off!

The deletion of `USER worker` was a test I made and it brought me to run docker compose successfully on top of not having any segfault error on mac (M2 chip), please do test it out and let me know the result.

Thanks again for engaging and keep up the good work!

PR #1435, opened 2023-12-20
Update Dockerfile.local

YasinZabun:
When i use as you added the dockerfile.local, i encountered "**Connection pool is full, discarding connection: pypi.org.Connection pool size : 10**" error. I solved the error by adding "**RUN poetry config installer.max-workers 10**" command to the dockerfile.


imartinez:
I got the same warning but didn't break the build.

PR #1437, opened 2023-12-21
Configurable context_window and tokenizer

imartinez:
Extract two key model parameters to settings. 
- Tokenizer: used to calculate the available context window given a prompt in several LlamaIndex pipelines. The default tokenizer is the one used for gpt-3.5 but it is not aligned with, for example, Mistral model. Using huggingface as provider of the tokenizer information, added to setup download script.
- context_window: base model argument

### DIFF
--- a/private_gpt/components/llm/llm_component.py
+++ b/private_gpt/components/llm/llm_component.py
@@ -1,11 +1,13 @@
 import logging
 
 from injector import inject, singleton
+from llama_index import set_global_tokenizer
 from llama_index.llms import MockLLM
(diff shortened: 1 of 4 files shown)


pabloogc (approved):

cognitivetech:
is there a way to make this setup script not download the model if I already have it?

running setup for embeddings \ tokenizer.. since I don't know how to add those manually but it also triggers superfluous model download.

alternatively it would be nice to have a documented way to manually download and configure embeddings \ tokenizer

Issue #1424, opened 2023-12-19
[Request] Allow custom openai endpoint


lenaxia:
Due to changes in PrivateGPT, openai replacements no longer work as we cannot define custom openai endpoints. 

LocalAI is an example: https://github.com/mudler/LocalAI/tree/master/examples/privateGPT

It is a drop in replacement for ChatGPT Openai with compatible REST APIs. Being able to define custom openai endpoints in the settings files would enable us to use localai.

creasydude:
any updates?

mchill:
This is pretty easy to implement with [OpenAILike](https://docs.llamaindex.ai/en/stable/api_reference/llms/openai_like.html). I tested it briefly at first with LocalAI, and then with vLLM. I also need this feature, so I'll get around to opening a PR if no one else picks it up.

PR #1447, opened 2023-12-22
feat(llm): Add openailike llm mode

mchill:
This mode behaves the same as the openai mode, except that it allows setting custom models not supported by OpenAI. It can be used with any tool that serves models from an OpenAI compatible API.

Note that the mentioned libraries, LocalAI and vLLM, also have their own providers in LlamaIndex. This is a more generic solution that should support those and many more libraries.

One fun implication that I'm testing is the possibility of running PrivateGPT on top of another instance of PrivateGPT. The top-level instance manages embeddings, and the bottom-level instance runs an LLM. With a little additional logic, it could be possible to route requests from the parent instance to multiple child instances based on model id in the request, effectively letting a single PrivateGPT instance serve multiple models while benefiting from a single document store.

Implements #1424

### DIFF
--- a/fern/docs/pages/manual/llms.mdx
+++ b/fern/docs/pages/manual/llms.mdx
@@ -37,6 +37,7 @@ llm:
   mode: openai
 
 openai:
+  api_base: <openai-api-base-url> # Defaults to https://api.openai.com/v1
   api_key: <your_openai_api_key>  # You could skip this configuration and use the OPENAI_API_KEY env var instead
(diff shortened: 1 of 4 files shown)


imartinez (approved):
I love the proposal and the execution. Great job. And the possibility of running a PGPT on top of another PGPT is really exciting. Thanks!

mchill:
Pushed a fix for the linting errors. I ran `make test` and completely skipped over the `make check` instruction. 😅

imartinez (approved):

Issue #1404, opened 2023-12-14
Modifications to use MOE models (Mixtral 8x7B instruct)


stutteringp0et:
Modifying 2 files is all you need.

pyproject.toml
llama-cpp-python = "^0.2.11" to llama-cpp-python = "^0.2.23"

poetry.lock
Search llama-cpp-python and update 2 values
version = "0.2.23"
    {file = "llama_cpp_python-0.2.23.tar.gz", hash = "sha256:364b61a13970932ea189b45a1c5dea89797b90e5da00f1fe6e72c47fbc512898"},

I only have one issue - the UI doesn't work anymore.  It loads, but any query crashes it.  The API seems to work fine though.

writinguaway:
Awesome! 

I did this, and then upgraded gradio and changing it in the same files, and was able to get results on the UI.

stutteringp0et:
@writinguaway - what gradio changes did you make?  Getting the UI working would be nice ;)

writinguaway:
Basically exactly the same as you did for llama-cpp-python, but with gradio. Off the top of my head: 

`pip install gradio --upgrade`
`vi poetry.lock' edit the 3x gradio lines to match the version just installed
'vi pyproject.py' edit the gradio line to match the version just installed

Then I was able to just run my project with no issues interacting with the UI as normal.

madjohn698:
@stutteringp0et ，@writinguaway,when I run 'make run', it display out of memory, my GPU is A2000(8G), RAM is 64G, and embedding_hf_model_name is BAAI/bge-small-en-v1.5, the llm_hf_model_file is mixtral-8x7b-instruct-v0.1.Q4_K_M.gguf,can you help me?

brooda:
Thould this be a setting for running this model?

local:
  prompt_style: "llama2"
  llm_hf_repo_id: TheBloke/Mixtral-8x7B-v0.1-GGUF
  llm_hf_model_file: mixtral-8x7b-instruct-v0.1.Q4_K_M.Q6_K.gguf
  embedding_hf_model_name: BAAI/bge-small-en-v1.5

stutteringp0et:
@brooda - not sure if it matters, but the llm_hf_repo_id for the mixtral instruct model is:
TheBloke/Mixtral-8x7B-Instruct-v0.1-GGUF

neofob:
The file `` needs to be updated as well to use with `llama-cpp-python=^0.2.23`.

[...]

PR #1479, opened 2024-01-03
Add support for Mixtral 8x7B model issue #1404

neofob:
Add support for Mixtral 8x7B model


imartinez:
While this PR may be useful for people to understand how to change the settings to use that model, it is not something we'd like to get into main branch. The reason is that that model is huge and most laptops won't be able to run it.

PR #1493, opened 2024-01-08
Skip file ingestion if already ingested

spirosbond:
Enhanced ingest_service.ingest_file() and ingest_service.bulk_ingest(…) to skip ingestion if the file(s) is already ingested.


darko-mijic:
It would be great for this to be reviewed and merged.

imartinez (changes requested):
You are relying on the existence of the metadata "file_name".
 d.doc_metadata["file_name"]
That is an assumption that should not be made in a core component such as IngestService.

Therefore I'd suggest to move all the logic to ingest_folder.py. That is a utility script, not part of the core codebase of PrivateGPT, specifically created to allow for local bulk ingestion. The same way it is ok to make metadata assumptions, for the same reason, in ui.py, you can use them in ingest_folder.py. 

Summing up, I'd move the filter_new_only, is_ingested, get_ingested and their usage in bulk_ingest to ingest_folder.py and not modify ingest_service.py

Robinsane:
Great find that ingesting same file happens, I just stumbled on this myself as well!

I agree with @imartinez to not do this in a core component. Users can request all ingested documents before ingesting new ones. However I do think the "Upload File(s)" button in the GUI should be adjusted to avoid this unintuitive behaviour:
1. Delete existing Documents with the same filename
2. Ingest the file as normal

negatives:
- slower (not a big difference as long as user doesn't mindlessly upload / inject same files)
positives:
+ most intuitive (based on what I saw in the GUI, I thought this was actually what happened)
+ handier for most real world use cases: File got updated? simply ingest it and it will overwrite any old one.
+ handier for non-technical people in the community focusing on the GUI

PR #1512, opened 2024-01-15
Add phi-2 support

regularfry:
This patch introduces the chatml prompt format (and associated tests) necessary for driving the [Dolphin 2.6](https://huggingface.co/cognitivecomputations/dolphin-2_6-phi-2) fine-tune of [`microsoft/phi-2`](https://huggingface.co/microsoft/phi-2).  I'm using the fine-tune rather than the upstream because the prompt behaviour with the chatml format seems to be more robust.

Also included is `settings-phi-2.yaml`, which you can use by setting `PGPT_PROFILES=phi-2` for a local experience that's tolerable on CPU.  On an M1 mac with Metal, I'm getting about 20 tokens per second output using the `Q4_K_M` model listed.

The `phi-2` architecture also needs the `llama-cpp-python` version to be updated, so I've bumped it to 0.2.28.


private_gpt/settings/settings_loader.py
```diff
@@ -16,7 +16,7 @@
 _settings_folder = os.environ.get("PGPT_SETTINGS_FOLDER", PROJECT_ROOT_PATH)
 
 # if running in unittest, use the test profile
-_test_profile = ["test"] if "tests.fixtures" in sys.modules else []
+_test_profile = ["test"] if "unittest" in sys.modules else []
```
imartinez, line 19: Don't revert this change. It was done on purpose to fix an issue with latests versions of llamaindex. Check d3acd85fe34030f8cfd7daf50b30c534087bdf2b

settings-phi-2.yaml
imartinez, line 1: Instead of adding another .yaml file, you could document it in https://docs.privategpt.dev/recipes/choice-of-llm/list-of-ll-ms. For that, just edit the fern/docs/pages/recipes/list-llm.mdx

imartinez (changes requested):
Great contribution!

private_gpt/settings/settings_loader.py
regularfry: Hm. That's odd. I don't remember making that change at all, there's no reason I would have needed to.  I wonder if I got something wrong when I committed the patch.  Easy enough to revert...

imartinez:
Closing it due to inactivity after latest comments

PR #1519, opened 2024-01-17
Fix: Adding an LLM param to fix generator producing only "#####" characters

naveenk2022:
The newer versions of `llama-cpp-python` do not have the parameter `offload_kqv` set to True. Fixing this in privateGPT is pretty straightforward. Instead of installing a lower version of `llama-cpp-python`, adding `"offload_kqv": True` to `model_kwargs` enables KGV offloading, and also significantly boosts GPU performance.


imartinez (approved):
Thanks for the fix!

PR #1532, opened 2024-01-23
feat: support reranker

Anhui-tqhuang:
@imartinez hey review appreciated

i want to add the support for reranker as a postnodeprocesser

The functionality of the `Reranker` is as follows:

1. It evaluates the similarity between a query and documents retrieved by the retriever.
2. If the similarity score is less than `cut_off`, the document is excluded from the results.
3. In scenarios where the filtered documents are fewer than `top_n`, the system defaults to providing the top `top_n` documents ignoring the `cut_off` score.
4. The `hf_model_name` parameter allows users to specify the particular FlagReranker model from [Hugging Face](https://huggingface.co/) for the reranking process.

Use the `enabled` flag to toggle the `Reranker` as per requirement for optimized results.


Anhui-tqhuang:
@imartinez hey review appirecated

imartinez:
@Anhui-tqhuang I think the idea is great! I have a general question though: those reranker models can be used through FlagEmebedding (as you proposed) or through SentenceTransformers. The benefit of the latter is that LlamaIndex already contains a SentenceTransformerRerank class. That way we could add the benefits you are proposing without:
- depending on an extra library (flagembedding)
- having to write a custom reranker class

I'd love to know your opinion. Maybe you want to give a try to the SentenceTransformer reranker using the same model. The rest of the PR looks like an amazing addition to the project. I'd ask you to update the documentation to reflect it (fern/docs/pages...)

Thanks!

Anhui-tqhuang:
Hey thanks for your suggestion! I would try to use the sentence transformer as well as updating the documentation when i back from the spring festival!

Anhui-tqhuang:
btw though llamax index has already supported the llm reranker. But it is not that stable since the output of llm is not expectable, as a result 
, private gpt might meet errors when the response from the llm cannot be parsed 

That's the reason i want to use a dedicated model which is used for reranker as its result will contains the similarity only

I might need some time on the investigation could we use the sentences transformer to run that dedicated model directly

Anhui-tqhuang:
@imartinez hey i give another pass to the model card again: https://huggingface.co/BAAI/bge-reranker-large#usage-for-reranker

it cannot be used with the sentence transformer

moreover with my previous comments, i still try to avoid using llm for reranker purpose as the result is not stable for `parser`, a dedicated model with `float` result would be more stable

Anhui-tqhuang:
@imartinez btw, i tried to update the doc, but the preview seems cannot work correctly with my PR

Anhui-tqhuang:
@imartinez this is doc https://privategpt-preview-c71c8bab-4dcb-4e83-ae34-1c22bb1ee919.docs.buildwithfern.com/manual/advanced-setup/reranker

Anhui-tqhuang:
@imartinez hey, could you plz have a pass? it will be appreciated

cloudrage999:
hi anhui
im using your PR in my pg isntance
im using bge-reranker as well
when you think this PR would be approved ?

also is bge-reranker the best local reranker we can have ?



another thing,i made a PR about query search results , so be able to do both semantic search and classic keyword-based search to retrieve more relevant context for user query

can you please take a look at it and tell me your opinion on it,its bugs and etc
thanks man

Anhui-tqhuang:
@cloudrage999 hey
> when you think this PR would be approved ?
i am still waiting for review from @imartinez 

> also is bge-reranker the best local reranker we can have ?
to be honest, i haven't tried other reranker models, so i cannot tell

Anhui-tqhuang:
@cloudrage999 
> another thing,i made a PR about query search results , in order be able to do both semantic search and classic keyword-?> > based search to retrieve more relevant context for user query

> can you please take a look at it and tell me your opinion on it,its bugs and etc
> thanks man

could you show me the link to the PR plz?

Issue #1559, opened 2024-01-30
How to delete ingested files from PGPT in DOCKER


CRPrinzler:
How is `make wipe ` implemented into the dockerized version of PGPT?
I have the container running and it works fine, but I cannot se how to invoke deletion of ingested context.

icsy7867:
You could use the API, or just use docker exec...

```
docker exec -it privategpt make wipe
```

Make sure to use whatever you named the container.  You can also just execute bash to get a shell.

CRPrinzler:
via docker gives : 

`OCI runtime exec failed: exec failed: unable to start container process: exec: "make": executable file not found in $PATH: unknown
`

icsy7867:
> via docker gives :
> 
> `OCI runtime exec failed: exec failed: unable to start container process: exec: "make": executable file not found in $PATH: unknown `

Whats your docker file look like?   Drop into a bash shell and make sure you are in the proper work directory:
```
docker exec -it privategpt bash
```

CRPrinzler:
I am, but it says /home/worker/app/

make not found

icsy7867:
install make into the container.

docker exec -it privategpt apt update && apt install make -y

Since the provided container is using layers, I dont think any of this stuff is installed perhaps.  If you can't use apt, you probably need to run as root:

```
docker run --rm
    --user=root privategpt bash
```

or something like that.

I actually re-wrote my docker file to just pull the github project in, as the original method seemed to be missing files.

CRPrinzler:
yes I am in the correkt folder:
`/home/worker/app`

I haven't got make or sudo.

CRPrinzler:
Essentially I just went in and rm -rf  under local_data . ..

PR #1568, opened 2024-02-01
Delete all files button

Robinsane:
Button to delete all ingested files. (+ ability to disable the button in settings.yaml)

When first using pgpt I could've used this myself (now I use Python SDK)
I saw multiple cases where people wanted something similar (issues / Discord).

issue example:
https://github.com/imartinez/privateGPT/issues/1559

### DIFF
--- a/private_gpt/settings/settings.py
+++ b/private_gpt/settings/settings.py
@@ -178,8 +178,11 @@ class UISettings(BaseModel):
     default_query_system_prompt: str = Field(
         None, description="The default system prompt to use for the query mode."
     )
-    delete_files_button_enabled: bool = Field(
-        True, description="If the button to delete all files is enabled or not."
+    delete_file_button_enabled: bool = Field(
+        True, description="If the button to delete a files is enabled or not."
+    )
+    delete_all_files_button_enabled: bool = Field(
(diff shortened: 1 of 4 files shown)


Robinsane:
Will look into expanding this PR /w button to delete just 1 file

Robinsane:
Added code with a button to delete a single file

Not happy yet with this:
- multiple users deleting single file at the same time => problem : only last selected file will be deleted (will not happen often & you can disable button in settings.yaml)
- I'd like to refresh the list of ingested files somehow but can't seem to figure out how

TobiasJu (commented):
Cloned the repo, ran the code. 
Thanks for your work!
Found one minor issue. The "Delete all files" button is permanently disabled.

image

Robinsane:
@TobiasJu 
This was intentional
See settings.yaml

`  delete_file_button_enabled: true
  delete_all_files_button_enabled: false`
  
  If most of you agree I'll make both true by default
  I was thinking for multiple users you don't want an accidental click to go poof (other button doesn't do anything unless a file is selected)

imartinez:
Included in https://github.com/imartinez/privateGPT/pull/1612

(btw, I made both buttons visible by default)

PR #1589, opened 2024-02-08
Updated ui.py to include a small sleep timer

icsy7867:
…m deltas.  This recursive function fires off so quickly to eats up too much of the CPU.  This small sleep frees up the CPU to not be bottlenecked.  This value can go lower/shorter.  But 0.02 or 0.025 seems to work well.

**Here you can view the slow down before the modifications (sorry for the poor quality):**

Screen Recording - End

**And here you can view the speed up after:**
Screen Recording 2024-02-07 at 6 30 29 PM

The entire process before this small change, the chat stream took 60-120 seconds or so.
**However after the change it takes about 8-10 seconds for me.**

### DIFF
--- a/private_gpt/ui/ui.py
+++ b/private_gpt/ui/ui.py
@@ -1,6 +1,7 @@
 """This file should be imported only and only if you want to run the UI locally."""
 import itertools
 import logging
+import time
 from collections.abc import Iterable
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


EEmlan (approved):
Have tested this.
Now the Generation in the UI runs significantly faster

ic-mw:
Excellent, it works nicely. I can confirm that the text output speed increases dramatically, about 10x. Good catch! 👍

lolo9538:
Thank you SO MUCH!!!!!

lolo9538:
Hi, I wonder if there is a similar issue for the ingest part. I have several thousands of documents to ingest and I see the GPU is not used at all and the CPU is barely used. I have upped the settings to 
embedding:
  mode: mock
  ingest_mode: parallel
  count_workers: 8

But the ingest is still very slow. Can anyone confirm whether ingest should be done by the GPU?
Thanks

imartinez (approved):

PR #1603, opened 2024-02-13
Update vector_store_component.py to add keyword based search alongside semantic search

cloudrage999:
hi,i tried to make a PR in order to increase the accuracy of search results ,im not sure if i did it in the right way but i tried 
the idea is to have both semantic search and traditional keyword based search in order to get more relevant results
i hope this feature be implemented in the project in future, maybe more modifications are needed

### DIFF
--- a/private_gpt/components/vector_store/vector_store_component.py
+++ b/private_gpt/components/vector_store/vector_store_component.py
@@ -1,18 +1,17 @@
 import logging
-import typing
-
 from injector import inject, singleton
 from llama_index import VectorStoreIndex
 from llama_index.indices.vector_store import VectorIndexRetriever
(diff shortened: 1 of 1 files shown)


semenov-d-v:
I want to highlight this PR, I think it is really important

jaluma (changes requested):
First of all, I would like to thank you for your PR but, currently, much of this code is unnecessary. The llama-index framework itself takes care of hybrid search if the provider supports it. I recommend looking at the `VectorStoreQueryMode` enum in the llama-index code.

If you can try, one first possible option to implement it would be to pass the `VectorStoreQueryMode` parameter to the Retriever. You would define a new setting to define which will be the selected query mode.

AlfonsoLozana:

PR #1618, opened 2024-02-18
set worker's home directory

optman:
some package depends on home directory to store cache data.  If don't set, the home directory will point to /nonexistent, then those packages may show permission deny warning. It also prevents vs-code to attach to the container.

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -34,7 +34,7 @@ ENV PORT=8080
 EXPOSE 8080
 
 # Prepare a non-root user
-RUN adduser --system worker
+RUN adduser --system --home /home/worker worker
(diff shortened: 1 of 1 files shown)


Dockerfile.local
```diff
@@ -34,7 +34,7 @@ ENV PORT=8080
 EXPOSE 8080
 
 # Prepare a non-root user
-RUN adduser --system worker
+RUN adduser --system --home /home/worker worker
```
qdm12, line 37: 👍 This is indeed important

Duplicated in my (larger) PR #1792

jaluma:
Closed. Merged https://github.com/zylon-ai/private-gpt/pull/1792

PR #1624, opened 2024-02-20
feat: support pgvector

Anhui-tqhuang:
add pgvector as one option to vector store with the following configurations

The available configuration options are:
| Field         | Description                                               |
|---------------|-----------------------------------------------------------|
| **host**      | The server hosting the Postgres database. Default is `localhost` |
| **port**      | The port on which the Postgres database is accessible. Default is `5432` |
| **database**  | The specific database to connect to. Default is `postgres` |
| **user**      | The username for database access. Default is `postgres` |
| **password**  | The password for database access. (Required)            |
| **embed_dim** | The dimensionality of the embedding model (Required)    |
| **schema_name** | The database schema to use. Default is `private_gpt`       |
| **table_name** | The database table to use. Default is `embeddings`    |


imartinez (dismissed):
Great addition and execution!

PR #1641, opened 2024-02-22
adding multiple GPU via CustomDataParallel

yukiman76:
adding adding multiple GPU via CustomDataParallel


yukiman76:
This PR Causes currency issues and is not recommend during ingesting data

PR #1643, opened 2024-02-23
Maintain score order when curating sources

Hoffelhas:
I noticed that every time after restarting PrivateGPT, the output on the screen would vary with similar queries used in "Search Files" mode even though the retrieved 'response' contained the same values that were ordered to the same scores.

After some troubleshooting, it looked like the output from the Source.curate_sources(response) function mixes up the values, which results in sources that are no longer ordered to the reponse scores.

In order to fix this, 'curated_sources' in the 'Source' class is changed from a set to a unique list to ensure that the score order is maintained and the results are consistent.


imartinez:
@Hoffelhas please run make check locally. It will fix format issues and check for typing errors.

Hoffelhas:
> @Hoffelhas please run make check locally. It will fix format issues and check for typing errors.

Done 👍

imartinez (approved):

Issue #1652, opened 2024-02-26
no CUDA devices found, CUDA will be disabled


Verzul13:
Hi! When I launch the project via docker compose, I get: no CUDA devices found, CUDA will be disabled. 
Here's what the command outputs: "nvcc --version":
`nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Tue_Aug_15_22:02:13_PDT_2023
Cuda compilation tools, release 12.2, V12.2.140
Build cuda_12.2.r12.2/compiler.33191640_0
`
nvidia-smi:
`+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.147.05   Driver Version: 525.147.05   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA GeForce ...  Off  | 00000000:03:00.0  On |                  N/A |
|  0%   40C    P8    17W / 200W |     18MiB / 12282MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
`
Command: docker run --rm --gpus all -it nvidia/cuda:12.2.2-devel-ubuntu22.04 bash

==========
== CUDA ==
==========

CUDA Version 12.2.2

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

I haven't tried anything, nothing helps, please help =(

Issue #1597, opened 2024-02-09
Docker GPU support permission error


Apotrox:
I have been sitting at this for 1.5 days now and i don't know where to go from here.
First i got it working with CPU inference by following imartez guide in #1445 and changing to this `docker compose run --rm --entrypoint="/usr/bin/env python3 scripts/setup" private-gpt`, after which i decided to take on gpu support.

The approach in #1405 seems to be the most promissing, but i can't get it to work. The Entrypoint always fails as it can't seem to find the entrypoint.sh file.

After that i tried around **a lot**. My most promissing candidate so far is taking the original dockerfile and modifying it with manually installing cuda toolkit for wsl like this:

```#install cuda toolkit
RUN wget https://developer.download.nvidia.com/compute/cuda/repos/wsl-ubuntu/x86_64/cuda-wsl-ubuntu.pin
RUN mv cuda-wsl-ubuntu.pin /etc/apt/preferences.d/cuda-repository-pin-600
#requires to have the file downloaded beforehand and put into the root folder
COPY cuda-repo-wsl-ubuntu-12-3-local_12.3.2-1_amd64.deb ./
RUN dpkg -i cuda-repo-wsl-ubuntu-12-3-local_12.3.2-1_amd64.deb
RUN cp /var/cuda-repo-wsl-ubuntu-12-3-local/cuda-*-keyring.gpg /usr/share/keyrings/
RUN apt-get update
RUN apt-get -y install cuda-toolkit-12-3
RUN rm cuda-repo-wsl-ubuntu-12-3-local_12.3.2-1_amd64.deb
ENV PATH="/usr/local/cuda-12.3/bin${PATH:+:${PATH}}"
```

and reinstalling llama-cpp-python with this (after poetry install local):

```ENV CUDA_HOME=/usr/local/cuda
# Enable GPU support
RUN CMAKE_ARGS='-DLLAMA_CUBLAS=on -DCUDAToolkit_ROOT=/usr/local/cuda/bin' poetry run pip install --force-reinstall --no-cache-dir llama-cpp-python
```

Which seems to get compiled and work. After a lot of headaches.

The last thing i'm running into is a permission error when trying compose and run this with `docker compose run --rm --entrypoint="/usr/bin/env python3 scripts/setup" private-gpt`:
`PermissionError: [Errno 13] Permission denied: '/home/worker/app/models/cache/models--intfloat--multilingual-e5-large/refs/main'`

[...]

Issue #1579, opened 2024-02-05
how do i change the models: llm or embedding


cleesmith:
It seems that editing settings.yaml and running again isn't enough to change to different models.

Are there any instructions for exactly how to do this?  I see nothing in the docs.

I'm ok if errors happen or the new models don't work properly.

Also, is it possible to use LLM Studio which provides OpenAI-like endpoint, as a way to more easily test other model?

Otherwise, this works well ... Thank you!

icsy7867:
Models have to be downloaded.

If you set the tokenizer model,  which llm you are using and the file name,  run scripts/setup and it will automatically grab the corresponding models.

Apotrox:
So far i changed the models (LLM and embedding) in the settings.yaml and settings-docker.yaml, and after running `docker compose run --rm --entrypoint="/usr/bin/env python3 scripts/setup" private-gpt` it seems to at least download them. it's still at it so i can't say for sure if it worked or not, will keep this updated.

JanuRam:
Was facing the same issue. The above solution solved it. Thank you @Apotrox

Walker555:
I wanted to use a German speaking model. Using Ubuntu with pyenv.

step 1- create settings-de.yaml
```
local:
  llm_hf_repo_id: TheBloke/em_german_leo_mistral-GGUF
  llm_hf_model_file:   em_german_leo_mistral.Q4_K_M.gguf
  embedding_hf_model_name: T-Systems-onsite/german-roberta-sentence-transformer-v2
  #llama, default or tag
  prompt_style: "default"
```
step 2 - download the model from huggingface hub with the setup script. It will keep all already downloaded files in models subdir.
```
PGPT_PROFILES=de ./scripts/setup
```
step 3 - run the app
```
PGPT_PROFILES=de python3.11 -m private_gpt
```

hamvi-ai:
is it possible to use LLM Studio which provides OpenAI-like endpoint, as a way to more easily test other model?

icsy7867:
Check out the openailike mode.

https://docs.privategpt.dev/manual/advanced-setup/llm-backends

terrornoize:
> I went into the settings-ollama.yaml and changed the name of the model there from Mistral to any other llama model. Whe nI restarted the Private GPT server it loaded the one I changed it to.

And directly download the model only with parameter change in the yaml file? Does the new model also maintain the possibility of ingesting personal documents?

Issue #1573, opened 2024-02-03
ValueError: Provided model path does not exist. Please check the path or provide a model_url to download.


dinubogdan1973:
Hello everyone,

I'm trying to install privateGPT and i'm stuck on the last command :

poetry run python -m private_gpt

I got the message "ValueError: Provided model path does not exist. Please check the path or provide a model_url to download."

Few details:

Python 3.11.5
pyenv 3.1.1
Poetry (version 1.7.1)

Someone can help me please ?

PS: I found that I have 2 folders in pypoetry/Cache/virtualenvs/:
my-project-0302-xxxxxxxx-py3.11 (named when poetry installed)
private-gpt-xxxxxxxx-py3.11

Shailendra420:
Facing same issue

Obfusgated:
same here

Apotrox:
you wanna run the scripts/setup before privategpt

ConstructionQA:
https://github.com/imartinez/privateGPT/issues/1303#issuecomment-1861976809

This helped to solve problem

Anhui-tqhuang:
```shell
$ poetry run python scripts/setup
```
run this first might help solve the problem

PR #1655, opened 2024-02-28
GPU support in Docker, other Docker-related updates

lukaboljevic:
The main contribution of this PR is the addition of a new Dockerfile and docker compose file for running PrivateGPT on GPU in Docker. The command to run is simply `docker compose -f docker-compose-gpu.yaml up --build`. This should address issues like #1652, #1597, #1405.

The PR also proposes some changes based on #1428:
- Added max workers for poetry installer
- Added an `entrypoint.sh` script. Here, the user can specify the model, tokenizer, prompt style and embedding model through environment variables, and then those are automatically downloaded using the setup script.
- Choose the tokenizer and prompt style in `settings-docker.yaml`, and update default Mistral model to v0.2 from v0.1
- Removed `USER worker` as it seems to have caused a segfault on Mac (can't test unfortunately, as I don't have a Mac)

I feel like having the `entrypoint.sh` script and simply running `docker compose up` is way simpler and more transparent than the current (and not very well documented) approach of running `docker compose run --rm --entrypoint="bash -c '[ -f scripts/setup ] && scripts/setup'" private-gpt`. The script also allows the user to more easily and directly choose model, tokenizer, prompt style and embedding model. The current approach involves creating a new `settings-FOO.yaml` file, and including FOO as a new profile in the corresponding docker compose file. This isn't bad at all, it just takes a while to find through the documentation, and may require a few attempts and browsing through issues like #1579 and #1573 before getting it to run.

Whatever your stance is on the entrypoint script, the documentation needs to be updated to tell the user as directly as possible what is the correct way to run and configure PrivateGPT in Docker. One shouldn't have to dig deep in the documentation and issues to look for something that already exists and works. I would really like to hear your opinion on this, so we can discuss on how exactly to update it.

In any case, I'm open for any comments and suggestions, and I hope you find this PR useful.

Edit: I ran `make test` and `make check` - 30 tests passed with 23 warnings, while make check fixed some files which I did not edit.

### DIFF
--- a/Dockerfile.local.gpu
+++ b/Dockerfile.local.gpu
@@ -1,4 +1,16 @@
-FROM python:3.11.6-slim-bookworm as base
+FROM nvidia/cuda:12.2.2-devel-ubuntu22.04 as base
+
+# For tzdata
+ENV DEBIAN_FRONTEND="noninteractive" TZ="Etc/UTC"
+
+# Install Python 3.11 and set it as default
+RUN apt-get update && \
+    apt-get install -y software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update && \ 
+    apt-get install -y python3.11 python3.11-venv python3-pip && \
+    ln -sf /usr/bin/python3.11 /usr/bin/python3 && \
+    python3 --version
 
 # Install poetry
 RUN pip install pipx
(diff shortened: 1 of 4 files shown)


neofob:
@lukaboljevic :
This is great! You beat me to it. I plan to work on this this weekend but you saved me some time. :)

What I would do is similar except a minor change where I would install `virtualenv` in the base (builder) stage, install deps..etc. then copy the virtual env directory to the `app` stage in Dockerfile. Anyhow, what you are doing is similar. I'll check out your PR and let you know how it goes.

neofob:
@lukaboljevic : The line 74 in Dockerfile.local.gpu should be
`ENV PYTHONPATH="$PYTHONPATH:/home/worker/app/private_gpt/"`

lukaboljevic:
Closed in favour of #1690 because of update to v0.4.0.

PR #1669, opened 2024-03-01
Check for duplicates in sources before outputting to the chat

icsy7867:
I found that after increasing
similarity_top_k: int  in vector_store_component.py to something higher, like 10, I would getting some duplicate hits in the sources used.  

Instead of just outputting the sources, as you loop through these, inject them into an array.  If the item already exists in the array, dont readd this to the sources_text.


icsy7867:
Whoops, I added the page to the check, and I forgot to update the if statement.  Will fix.

icsy7867:
Fixed with:
https://github.com/imartinez/privateGPT/pull/1669/commits/3d092ddffb2691384a2978ddc52ecaef2813cf23

PR #1703, opened 2024-03-11
Recreated settings changes - Adds serveral options for llamacpp and ollama

icsy7867:
Original PR here:
https://github.com/imartinez/privateGPT/pull/1677

llama-cpp https://llama-cpp-python.readthedocs.io/en/latest/api-reference/
https://docs.llamaindex.ai/en/stable/examples/llm/llama_2_llama_cpp.html#

ollama - https://github.com/run-llama/llama_index/blob/eeb2a60387b8ae1994005ad0eebb672ee02074ff/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py

No configurable changes. -
openailike - https://docs.llamaindex.ai/en/stable/examples/llm/localai.html#localai

Not sure about the model_kwargs. The value is references for openai, but I could not find documentation on what values were allowed.
openai - https://github.com/run-llama/llama_index/blob/eeb2a60387b8ae1994005ad0eebb672ee02074ff/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
https://docs.llamaindex.ai/en/stable/examples/llm/openai.html

For the text/description I used the values found here:
https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values

LlamaCPP, where it used the same K/V, had the same values. However my setup is currently using ollama, need some testing done for LlamaCPP.

I also added the temperature under the main llm.settings. This should allow the models that supports this value to be edited/changed.

### DIFF
--- a/private_gpt/__init__.py
+++ b/private_gpt/__init__.py
@@ -1,4 +1,5 @@
 """private-gpt."""
+
 import logging
 import os
 
--- a/private_gpt/components/llm/llm_component.py
+++ b/private_gpt/components/llm/llm_component.py
@@ -39,16 +39,23 @@ class LLMComponent:
                     ) from e
 
                 prompt_style = get_prompt_style(settings.llamacpp.prompt_style)
-
+                settings_kwargs = {
+                    "tfs_z": settings.llamacpp.tfs_z,  # ollama and llama-cpp
+                    "top_k": settings.llamacpp.top_k,  # ollama and llama-cpp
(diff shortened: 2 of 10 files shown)


icsy7867:
Hmm small bug... num_predict: 128 doesnt do what I think it does.  It tells llamaindex the maximum size of the response.  So this should probably be set to -1 or -2 by default.

It is odd though, that the default says "128", but if you dont set that kwarg, you get a larger response.

imartinez (approved):

PR #1706, opened 2024-03-11
Adding Postgres for the doc and index store

dbzoo:
Adding Postgres support for the Doc and Index store.
Provided example using ollama and everything stored in Postgres.

One thing that is a little unusual is that if you are using the same postgres DB for both the Docstore and the Vector store
you need to supply the credentials twice.  This is has both benefit and downsides.


private_gpt/components/node_store/node_store_component.py
```diff
@@ -16,19 +17,38 @@ class NodeStoreComponent:
...
-        try:
-            self.doc_store = SimpleDocumentStore.from_persist_dir(
-                persist_dir=str(local_data_path)
-            )
-        except FileNotFoundError:
-            logger.debug("Local document store not found, creating a new one")
-            self.doc_store = SimpleDocumentStore()
+    def __init__(self, settings: Settings) -> None:
+        match settings.docstore.database:
+            case "local":
```
imartinez, line 22: I'd call it "simple" instead of "local" to better reflect the option being selected, given postgres can also be local

private_gpt/components/node_store/node_store_component.py
```diff
@@ -16,19 +17,38 @@ class NodeStoreComponent:
...
+                try:
+                    self.doc_store = SimpleDocumentStore.from_persist_dir(
+                        persist_dir=str(local_data_path)
+                    )
+                except FileNotFoundError:
+                    logger.debug("Local document store not found, creating a new one")
+                    self.doc_store = SimpleDocumentStore()
+
+            case "postgres":
+                from llama_index.core.storage.index_store.postgres_index_store import PostgresIndexStore
```
imartinez, line 40: This needs to be wrapped in a try - catch, because otherwise PrivateGPT would not work if the dependency is not installed. Check vector store component for example. Something like this:

```
                try:
                    from llama_index.vector_stores.postgres import (  # type: ignore
                        PGVectorStore,
                    )
                except ImportError as e:
                    raise ImportError(
                        "Postgres dependencies not found, install with `poetry install --extras vector-stores-postgres`"
                    ) from e
```

imartinez (changes requested):
Very good addition! Added a couple minor comments. 

Also, an important one: we need to update pyproject.toml to include as optional
llama-index-storage-docstore-postgres and llama-index-storage-index-store-postgres; with that update the installation docs, etc. I can help with that if needed.

It'd be great to add specific documentation about this addition, the same way we already have a section about vector stores at https://docs.privategpt.dev/manual/storage/vector-stores

private_gpt/components/node_store/node_store_component.py
dbzoo: I like simple better than local too.  You are right, this better reflects the underlying storage type.

private_gpt/components/node_store/node_store_component.py
dbzoo: postgres is a subset of the pgvector requirements.  To keep things simple, I'll create the dependency to vector-stores-postgres extra rather than introduce another.

imartinez: I don't fully follow. I'd separate dependencies as much as possible, to allow for example using Postgres for index and doc store but not as vector store.

dbzoo: You'll note that I have created a single dependency ```storage-postgres``` for index/docstore postgres supporta.   This is due to the fact that both of these are used together, but they are separated in llama-index.  Also I have created dependencies on two PG libraries.  This is because the llama-index postgres component don't do it.  They should.  I need to raise a pull against llama-index to fix that. To prevent this project failing I've included them.

settings.yaml
```diff
@@ -57,7 +57,7 @@ vectorstore:
   database: qdrant
 
 docstore:
```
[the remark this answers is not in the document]
dbzoo: I'm glad you mentioned this.  I was mulling the setting over in my head after I committed the change, and I also felt there was an inconsistency issue as both index/docstore are treated as single storage setting.   Node is a much better fit.

pyproject.toml
```diff
@@ -27,6 +27,12 @@ llama-index-embeddings-openai = {version ="^0.1.6", optional = true}
 llama-index-vector-stores-qdrant = {version ="^0.1.3", optional = true}
 llama-index-vector-stores-chroma = {version ="^0.1.4", optional = true}
 llama-index-vector-stores-postgres = {version ="^0.1.2", optional = true}
+llama-index-storage-docstore-postgres = {version ="^0.1.2", optional = true}
+llama-index-storage-index-store-postgres = {version ="^0.1.2", optional = true}
+# Postgres
+psycopg2-binary = {version ="^2.9.9", optional = true}
```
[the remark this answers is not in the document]
dbzoo: I'm glad you raised this.  Both [docstore](https://github.com/run-llama/llama_index/blob/6cd92affa5835aa21f823ff985a81f006c496bbd/llama-index-integrations/storage/docstore/llama-index-storage-docstore-postgres/pyproject.toml#L32) and [index](https://github.com/run-llama/llama_index/blob/6cd92affa5835aa21f823ff985a81f006c496bbd/llama-index-integrations/storage/index_store/llama-index-storage-index-store-postgres/pyproject.toml#L32) don't define pg dependencies in their poetry configuration file.
Contrast to [pgvector](https://github.com/run-llama/llama_index/blob/6cd92affa5835aa21f823ff985a81f006c496bbd/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/pyproject.toml#L32) that does define PG library dependencies. 
I found that without explicitly mentioning them as dependencies, the would not get brought in.
 Personally, I think the defect is on llama_index, it should include the PG dependencies on those modules just like pgvector does in its poetry file.  When llama-index fixes this issue (or I create a PR) then we can remove them from our file.  Without them, the following won't bring in postgres libraries unless you also happen to be using pgvector.
` poetry install --extra storage-postgres`

HenrikPedDK:
Hi guys. I have this version running. There is some typos in the doc and a syntax error in the node_component.py but i got it up and running. Something is very wrong when querying files compared to before. I put an example in a thread in Discord. I don't know if it's the embedding that goes wrong or something else, but if I embed the same document directly in the running Ollama instance i get much better answers (same model / document)

dbzoo:
> There is some typos in the doc and a syntax error in the node_component.py

Those should both be fixed now.   What discord thread?

imartinez:
> > There is some typos in the doc and a syntax error in the node_component.py
> 
> Those should both be fixed now. What discord thread?

Summing up: @HenrikPedDK found that using Postgres for doc and index store makes the Query Docs not work properly. I haven't tested it myself, so it is not a 100% confirmed issue (at least not the direct relation with these changes).

Copying @HenrikPedDK feedback here from Discord (moving the convo to Discord is not a good practice; having said that, happy to also chat in [Discord](https://discord.com/invite/bK6mRVpErU) ):

Okay, i have ingested a cook book that i usually test with. Before using Postgres as index/docstore i could query for a specific recipe. It can't do that anymore, it summarizes the book okay but when asking into specific things it gives answers like this:

"I apologize for any confusion earlier, but I cannot directly provide you with the Cassoulet recipe from "Cast Iron Cooking For Dummies" as I don't have access to the actual text of the book. However, I can give you a basic outline and ingredients list for making a traditional French Cassoulet in a cast iron Dutch oven or casserole dish."
If i do the same directly on the same Ollama instance with the same document/mistral i get this answer:

Mistral:Latest 12/03/2024 13:52
Yes, I can provide you with a recipe for Cassoulet. Here it is:


INGREDIENTS



4 tablespoons olive oil

½ pound haricot coverts (green beans), cleaned

1 shallot, sliced

2 cloves garlic, sliced

1 tablespoon butter

¼ cup almonds, sliced

Salt and pepper to taste

Red pepper flakes to taste

1 pound lamb shoulder, medium diced

1 pound pork shoulder, medium diced

Salt and pepper, to taste

2 onions, small diced

4 garlic cloves, minced

2 tablespoons tomato paste

4 Roma tomatoes, small diced

1 teaspoon sugar

1 pound dry white beans, great northern, presoaked overnight

2 quarts chicken stock

4 sprigs thyme

2 bay leaves

3 sprigs parsley, separate leaves (mince) from stems (reserve)

3 slices bacon, medium diced

2 cups panko breadcrumbs

1 lemon zest


COOK TIME: 2 hours PREP TIME: 10 minutes YIELD: 8 servings


Heat the oil in a Dutch oven or large pot. Add the green beans, shallot, and garlic. Cook until tender, about 5 minutes. Remove from the pot and set aside.


Preheat the oven to 375 degrees Fahrenheit.


In the same pot, heat the butter. When it's melted, add the almonds. Cook for 2 minutes, then remove them and set aside with the green beans.


Season the lamb and pork with salt and pepper to taste. In the same pot, cook the meats until browned on all sides. Remove and set aside with the green beans and almonds.


In the same pot, heat 2 tablespoons oil. Add the onions and garlic. Cook for 3 minutes, then remove them and set aside.


Heat 1 more tablepoon of oil in the same pot. Add the tomato paste. Cook for 2 minutes. Then add the Roma tomatoes, sugar, and cooked white beans. Stir well to combine, then add chicken stock, thyme sprigs, bay leaves, and parsley stems.

( i removed the rest)

imartinez (changes requested):
Super small doc typos

private_gpt/components/node_store/node_store_component.py
```diff
@@ -16,19 +17,49 @@ class NodeStoreComponent:
...
+                    **settings.postgres.model_dump(exclude_none=True)
+                )
+                self.doc_store = PostgresDocumentStore.from_params(
+                    **settings.postgres.model_dump(exclude_none=True)
+                )
+
+            case _:
+                # Should be unreachable
+                # The settings validator should have caught this
+                raise ValueError(f"Database {settings.nodestore.database} not supported")
```
imartinez, line 65: This will fix the current test error:
Change
```
                raise ValueError(f"Database {settings.nodestore.database} not supported")
```
to
```
raise ValueError(
                    f"Database {settings.nodestore.database} not supported"
                )
````

Comment (imartinez) [reply private_gpt/components/node_store/node_store_component.py:65]: Screenshot 2024-03-13 at 10 04 13

**Review by imartinez — CHANGES_REQUESTED** [review]
Added a comment to fix the only pending linting error. 

Forget about the fern error, I'll disable that check in the near future

File: private_gpt/components/node_store/node_store_component.py (hunk 5 above)
Comment (dbzoo) [reply private_gpt/components/node_store/node_store_component.py]: Thanks - RUFF and BLACK are new to me.  I'm coming up to speed quickly with these new tools.  Future pull won't have these issues now I understand the process.

File: pyproject.toml [hunk 6]
```diff
@@ -46,7 +52,7 @@ embeddings-sagemaker = ["boto3"]
 vector-stores-qdrant = ["llama-index-vector-stores-qdrant"]
 vector-stores-chroma = ["llama-index-vector-stores-chroma"]
 vector-stores-postgres = ["llama-index-vector-stores-postgres"]
-
+storage-postgres = ["llama-index-storage-docstore-postgres","llama-index-storage-index-store-postgres","psycopg2-binary","asyncpg"]
```
dbzoo, line 55: I was thinking last night.   Might we want to call this `nodestore-postgres` for consistency too?

imartinez: We've tried to stick with LlamaIndex notation: 
llama-index-vector-stores-chroma -> vector-stores-postgres

In this case, given we are aggregating both docstore and index-store, I think the most correct way would be to call it `storage-nodestore-postgres` (given "lama-index-storage-docstore-postgres","llama-index-storage-index-store-postgres)

imartinez (approved):

imartinez (approved):
Great contribution and execution!

PR #1712, opened 2024-03-12
Feat/mongo

paul-asvb:


dbzoo:
This mongo PR is going to collide with https://github.com/imartinez/privateGPT/pull/1706
I note that you are only putting the index into mongoDB.   I played around with mongoDB.  The maximum size of a JSON object, caused me no end of pain, why I moved to Postgres, thus my PR.

Do we want to have separate index/docstore or a unified nodestore ?  Unification is direction of my PR.
simple is unified, postgres is unified.

PR #1730, opened 2024-03-15
Unify settings for vector and nodestore connections to PostgreSQL

dbzoo:
The duel configuration of the **postgres** connection settings for the **nodestore** and the **vector store** has a smell about it.   This PR unifies the connection to a **postgres** database by changing the database mode string from **pgvector** to **postgres**. 

It has an opinionated configuration for the name of the embedding table, and removes it from the configuration.  As the vector table needs to know the embedding model dimension to correctly setup the table, this has been moved into the `embedding:` section.

It was easier to create a PR than trying to explain all this in a discussion forum.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,6 @@ __pycache__/
 .vscode/
 /.run/
 .fleet/
-*~
-*brett*
 
 # macOS
 .DS_Store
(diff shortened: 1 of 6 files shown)


imartinez (approved):
Makes total sense. Thanks for bringing this improvement. Makes things cleaner and more scalable.

PR #1734, opened 2024-03-15
Updated to allow the selection of GPU for embedding where there is mo…

sgresham:
Updated to allow the selection of GPU for embedding where there is more than one available. Defaults to cuda[0] or cpu if cuda is not available. Commented reference in settings.yaml under embedding.

### DIFF
--- a/private_gpt/components/embedding/embedding_component.py
+++ b/private_gpt/components/embedding/embedding_component.py
@@ -7,7 +7,7 @@ from private_gpt.paths import models_cache_path
 from private_gpt.settings.settings import Settings
 
 logger = logging.getLogger(__name__)
-
+import torch
 
 @singleton
 class EmbeddingComponent:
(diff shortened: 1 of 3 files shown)


private_gpt/components/embedding/embedding_component.py
```diff
@@ -7,7 +7,7 @@
 from private_gpt.settings.settings import Settings
 
 logger = logging.getLogger(__name__)
-
+import torch
```
imartinez, line 10: I'd move this to the `try` block within "huggingface" case. There is no "torch" general dependency declared in pyproject.toml, so this could break the whole execution for people not using huggingface. Actually, we may need to add torch to `embeddings-huggingface = ["llama-index-embeddings-huggingface"]` as
```
# Optional Huggingface related dependency
torch = {version = "^2.2.1", optional = true}

embeddings-huggingface = ["torch", "llama-index-embeddings-huggingface"] 
```
in pyproject.toml. 

I think huggingface package from llamaindex already depends on torch, but given we are now importing it explicitly we should also depende on it.

private_gpt/components/embedding/embedding_component.py
```diff
@@ -28,9 +28,33 @@ def __init__(self, settings: Settings) -> None:
...
+                    # If settings.embedding.gpu is specified, use that GPU index
+                    if hasattr(settings, 'embedding') and hasattr(settings.embedding, 'gpu'):
+                        gpu_index = settings.embedding.gpu
+                        device = torch.device(f"cuda:{gpu_index}")
+                    else:
+                        # Use the default GPU (index 0)
+                        device = torch.device("cuda:0")
+                else:
+                    # If CUDA is not available, use CPU
+                    device = torch.device("cpu")
```
imartinez, line 52: What happens with laptops using a GPU that is not Nvidia based? For example Mac book running Metal GPU? Will this make embedding slower forcing them to go to CPU?

settings.yaml
```diff
@@ -54,6 +54,7 @@ embedding:
   # Should be matching the value above in most cases
   mode: huggingface
   ingest_mode: simple
+  # gpu: cuda[0]      # if you have more than one GPU and you want to select another. defaults to cuda[0], or cpu if cuda not available
```
imartinez, line 57: You'd need to include this new setting to `settings.py`

private_gpt/components/embedding/embedding_component.py
```diff
@@ -28,9 +28,33 @@ def __init__(self, settings: Settings) -> None:
                         "Local dependencies not found, install with `poetry install --extras embeddings-huggingface`"
                     ) from e
 
+                # Get the number of available GPUs
+                num_gpus = torch.cuda.device_count()
```
imartinez, line 32: Adding code to the codebase just to print information is not a good practive. I'd remove this. whole block of prints.

imartinez (changes requested):
I'm not so sure about this one. May be too specific to Nvidia setups.

private_gpt/components/embedding/embedding_component.py
dbzoo: This logic looks similar this: [llama_index.core.utils.infer_torch_device](https://github.com/run-llama/llama_index/blob/5c53f41712785e5558156372bdc4f33a6326fa5f/llama-index-core/llama_index/core/utils.py#L463) which handles Metal (mps).

AlfonsoLozana:

PR #1758, opened 2024-03-18
Fixed docker-compose

mrepetto-certx:
Make the compose work avoiding Pydantic error not recognizing local as proper mode.


mrepetto-certx:
My bad should be `llamacpp` and not `ollama`.

imartinez (approved):

PR #1776, opened 2024-03-21
Add nbconvert to poetry

ppinkhasov:
The environment can't add notebook files in the UI without nbconvert.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,8 @@ pytest-cov = "^3"
 ruff = "^0"
 pytest-asyncio = "^0.21.1"
 types-pyyaml = "^6.0.12.12"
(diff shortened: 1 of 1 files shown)


imartinez:
Could you elaborate? We are trying to reduce the dependencies as much as possible, and focus on the ones needed to deploy PrivateGPT to production setups. Notebooks are not the priority. Maybe a note in the documentation or an optional dependency would be better.

ppinkhasov:
No worries, just noticed the error by default poetry env that it spits out an error even trying to upload the notebook files. This fixes that issue.

imartinez:
Trying to keep dependencies to the minimum, closing this PR as notebook files are not generally used as an input.

PR #1785, opened 2024-03-22
replaced concat of strings

igeni:
replaced concat of strings to avoid potential mismatch of types and make code more clear


qdm12:
> replaced concat of strings to avoid potential mismatch of types

Why would there be a mismatch of types though? 🤔

imartinez (approved):
No major impact but definitely a source code improvement. Thanks!

PR #1812, opened 2024-03-31
Fixed Docker version with Ollama

mrepetto-certx:
Given that we have Ollama as the new recommended setup, I propose to make a modification to the Dockerfile and compose it to make it coherent with the new local setup. I decoupled Ollama from privateGPT, making it a separate microservice.


imartinez (approved):
Makes total sense, thanks!

PR #1822, opened 2024-04-02
Updated prompt_style to be included under the LLM setting and not jus…

icsy7867:
prompt_style was only under LlamaCPP. I have implemented this with openailike, though I think all of the llamaindex intergrations can use it.  I have placed prompt_style under llm, which can easily be added to others after testing.

I also included the token limit and temperature settings for openailike as well.


icsy7867:
Closing this, as I see my pull was a bit old and missing some newer code.  Will recreate.

PR #1825, opened 2024-04-03
Display ingested file paths instead of file names

omartrigui:
## Description

The current implementation only displays the ingested filenames, which can be confusing for users when the specified directory for ingestion contains nested directories with multiple levels. Additionally, the existing implementation discards files with different names e.g. a root_path directory contains index.md, dir1/index.md, dir2/index.md, only one single `index.md` file is displayed. This PR enhances the display by including the file paths instead of just filenames, addressing the potential confusion caused by nested directories, and ensures that all files are appropriately displayed without discarding any.

## Fix preview

Screenshot from 2024-04-03 10-10-22

### DIFF
--- a/private_gpt/components/ingest/ingest_helper.py
+++ b/private_gpt/components/ingest/ingest_helper.py
@@ -73,7 +73,7 @@ class IngestionHelper:
     ) -> list[Document]:
         documents = IngestionHelper._load_file_to_documents(file_name, file_data)
         for document in documents:
-            document.metadata["file_name"] = file_name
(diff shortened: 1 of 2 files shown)


icsy7867:
Personally this would break my use case for PGPT for me. Almost all of my items are web URLS.  So the "filename" is actually the URL of a confluence page, web article, policy, etc...

This PR would not work with how I use PGPT.  Not sure how others feel, but I think this should be a toggleable (Off by default) setting. 

IMO.

AlfonsoLozana:

PR #1841, opened 2024-04-10
Fixed "no such group" error in Dockerfile, added docx2txt and cryptography deps

dividebysandwich:
Creating a docker container was giving me a "No such group" error for the worker group. This has been fixed.
Also, I have added docx2txt to poetry, so users no longer get the error that docx2txt is not installed when running privategpt in a docker container.
cryptography has been added as well to remove the error when trying to parse certain PDF files.


imartinez (approved):

pyproject.toml
```diff
@@ -13,6 +13,8 @@ injector = "^0.21.0"
 pyyaml = "^6.0.1"
 watchdog = "^4.0.0"
 transformers = "^4.38.2"
+docx2txt = "^0.8"
+cryptography = "^3.1"
```
drupol, line 17: This new line is breaking local installation:

```
       > Checking runtime dependencies for private_gpt-0.5.0-py3-none-any.whl
       >   - cryptography<4.0,>=3.1 not satisfied by version 42.0.5
```

FYI, the current [`cryptography`](https://pypi.org/project/cryptography/) package is at version 42.

dividebysandwich: I did not try it with version 42. I used 3.1 because that was specifically requested by other dependencies. I can give that a go tomorrow. But I tested both local installs and Docker on a blank VM and both worked. Still, using a more current version is probably better so I'll check if that works, of if the newest version causes any problems.

drupol: Cool, looking forward for that!

dividebysandwich: > Cool, looking forward for that!

Appears to work fine.
See https://github.com/dividebysandwich/private-gpt/commit/ef516456829a2bdff02746bc074abcda906e3605 - but I haven't tested very thoroughly. Please let me know if this solves your problem.

PR #1878, opened 2024-04-23
 Model Temperature editing enabled on UI

Castolus:
Model Temperature editing enabled on UI. This option enables the user to "play" with the model´s spontaneity without having to restart the code every time. 
Visualization enabled from "settings.yaml". Range of Temperature goes from 0.05 to 0.95 since 0 and 1 are not possible values for this parameter. 
I didn´t add the rest of parameters (top_k, top_p, etc..)  as i belive it´d bee too much info in the UI. But I´m open to do it if you consider it´d be nice.

### DIFF
--- a/private_gpt/settings/settings.py
+++ b/private_gpt/settings/settings.py
@@ -151,6 +151,10 @@ class HuggingFaceSettings(BaseModel):
     embedding_hf_model_name: str = Field(
         description="Name of the HuggingFace model to use for embeddings"
     )
+    access_token: str = Field(
+        None,
+        description="Huggingface access token, required to download some models",
+    )
 
 
 class EmbeddingSettings(BaseModel):
--- a/private_gpt/settings/settings.py
+++ b/private_gpt/settings/settings.py
@@ -151,10 +151,6 @@ class HuggingFaceSettings(BaseModel):
     embedding_hf_model_name: str = Field(
         description="Name of the HuggingFace model to use for embeddings"
     )
-    access_token: str = Field(
-        None,
-        description="Huggingface access token, required to download some models",
-    )
 
 
 class EmbeddingSettings(BaseModel):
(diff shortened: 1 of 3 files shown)


private_gpt/settings/settings.py
```diff
@@ -151,10 +151,6 @@ class HuggingFaceSettings(BaseModel):
     embedding_hf_model_name: str = Field(
         description="Name of the HuggingFace model to use for embeddings"
     )
-    access_token: str = Field(
```
imartinez, line 154: Please don't remove this

imartinez (changes requested):
Good addition. Please check my comments to avoid breaking a prev functionality.

private_gpt/ui/ui.py
```diff
@@ -206,6 +206,10 @@ def _set_system_prompt(self, system_prompt_input: str) -> None:
         logger.info(f"Setting system prompt to: {system_prompt_input}")
         self._system_prompt = system_prompt_input
 
+    def _set_model_temperature(self, model_temperature_input: str) -> None:
+        logger.info(f"Setting Model Temperature to: {model_temperature_input}")
+        self._model_temperature = model_temperature_input
```
jaluma, line 211: Where do you use `self.model_temperature`? You are setting the property, but not sending it to the server. Are there any changes missing? @Castolus

private_gpt/ui/ui.py
```diff
@@ -367,6 +371,15 @@ def _build_ui_blocks(self) -> gr.Blocks:
...
                     )
+                    adjust_model_temperature = gr.components.Number(
+                        value=settings().llm.temperature,
+                        label="Model Temperature",
+                        interactive=True,
+                        visible=settings().ui.adjust_temperature_enabled,
+                        minimum=0.05,
+                        maximum=0.95,
+                    )
+
```
jaluma, line 382: I would suggest adding a button to be clearer. Right now, we have to press Enter to accept the change. It doesn't seem to me the most accessible function.

jaluma (changes requested):

Issue #1889, opened 2024-04-26
Please add Llama3 Prompt support thanks


psychopatz:
I want to use the newest Llama 3 model for the RAG but since the llama prompt is different from mistral and other prompt, it doesnt stop producing results when using the Local method, I'm aware that ollama has it fixed but its kinda slow to my liking than the local methon>

I made a dirty fix template to the prompt_helper.py by following the meta docs on prompting https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-3/ 

#prompt_helper.py Code
class Llama3PromptStyle(AbstractPromptStyle):
    def _messages_to_prompt(self, messages: Sequence[ChatMessage]) -> str:
        prompt = "<|begin_of_text|>"
        for message in messages:
            role = message.role
            content = message.content or ""
            prompt += f"<|start_header_id|>{role.lower()}<|end_header_id|>"
            prompt += f"{content.strip()}<|eot_id|>"
        prompt += "<|end_of_text|>"
        return prompt

    def _completion_to_prompt(self, completion: str) -> str:
        return self._messages_to_prompt(
            [ChatMessage(content=completion, role=MessageRole.USER)]
        )

def get_prompt_style(
    prompt_style: Literal["default", "llama2","llama3", "tag", "mistral", "chatml"] | None) -> AbstractPromptStyle:
    """Get the prompt style to use from the given string.

    :param prompt_style: The prompt style to use.
    :return: The prompt style to use.
    """
    if prompt_style is None or prompt_style == "default":
        return DefaultPromptStyle()
    elif prompt_style == "llama2":
        return Llama2PromptStyle()
    elif prompt_style == "llama3":  //Here's the changes, in theory it should trigger
        return Llama3PromptStyle()
    elif prompt_style == "tag":
        return TagPromptStyle()
    elif prompt_style == "mistral":
        return MistralPromptStyle()
    elif prompt_style == "chatml":
        return ChatMLPromptStyle()
    
    raise ValueError(f"Unknown prompt_style='{prompt_style}'")
    
    
 # Console Error
llamacpp.prompt_style
  Input should be 'default', 'llama2', 'tag', 'mistral' or 'chatml' [type=literal_error, input_value='llama3', input_type=str]

Regardless of it, I dont understand why it still produces an error. Sorry I'm new to these stuff.
I hope privategpt uses a template system similar to other LLM wrapper for added flexibility

Thanks in advance privategpt team

neofob:
[...]

PR #1962, opened 2024-06-06
added llama3 prompt

hirschrobert:
I was picking up issue #1889 , but had to change it a bit. meta documentation is here https://llama.meta.com/docs/model-cards-and-prompt-formats/meta-llama-3/.
Also some issues with double BOS tokens, so no BOS tokens here, see https://github.com/abetlen/llama-cpp-python/pull/1439.


private_gpt/components/llm/prompt_helper.py
jaluma, line 1: It would be great to have a unit test to test the output as well as the rest of the `PromptStyles`.

jaluma:
> Hey @imartinez , is it possible to change the pull request rules so that workflows are run without approval. What is the reason for it? Without approval, tests can be run already and code improved by contributors beforehand so that merging by a maintainer should only be a formal issue. This should also speed up things. Many thanks!

We have just changed how workflows are executed. Hopefully there won't be any more problems in the future :)

pyproject.toml
jaluma, line 1: This change remove a warning calling `lint`...

jaluma:
Updated PR with requested changes!

imartinez (approved):
👏 🦙

jaluma (approved):

PR #1963, opened 2024-06-06
Fixes

hirschrobert:
see commits


private_gpt/settings/settings.py
```diff
@@ -120,6 +120,10 @@ class LLMSettings(BaseModel):
 
 class VectorstoreSettings(BaseModel):
     database: Literal["chroma", "qdrant", "postgres"]
+    inject_win_size: int = Field(
+        3,
+        description="How many sentences on either side to capture, when parsing files",
+    )
```
jaluma, line 126: Can you move this property to another setting? It would be nice if we have different providers like graphs, elastic search, etc.

settings.yaml
```diff
@@ -74,6 +74,7 @@ huggingface:
 
 vectorstore:
   database: qdrant
+  inject_win_size: 2
```
jaluma, line 77: Please remove this line or set default value

jaluma (changes requested):

AlfonsoLozana:

PR #1965, opened 2024-06-07
Support for Google Gemini LLMs and Embeddings

uw4:
Initial support for Gemini, enables usage of Google LLMs and embedding models (see settings-gemini.yaml)

Install via
poetry install --extras "llms-gemini embeddings-gemini"

Notes:
* had to bump llama-index-core to later version that supports Gemini
* poetry --no-update did not work: Gemini/llama_index seem to require more (transient) updates to make it work...


raphant:
Gemini's 1m context window for pro and flash will seem to be a great aid to ingesting and analyzing big documents. Looking forward to seeing this implemented!

pyproject.toml
```diff
@@ -47,6 +49,9 @@ sentence-transformers = {version ="^2.6.1", optional = true}
 # Optional UI
 gradio = {version ="^4.19.2", optional = true}
 
+# Optional Google Gemini dependency
+google-generativeai = {version ="^0.5.4", optional = true}
```
jaluma, line 53: Why is it necessary to specify this package? It is already included in the `llama-index-llms-gemini` and `llama-index-embeddings-gemini`

jaluma: Also, can you fix mypy problems related to bump `llama-index`?

jaluma (approved):

jaluma:
Update PR with last changes

PR #1972, opened 2024-06-14
Dockerfile for a image that can be used in a pod on kubernetes

Kraego:
Kubernetes assigns a random user id to the pod, so the `chown` to worker in the present Dockerfiles won't be sufficient. In the new docker file the ownership is handeled with a group.

I've tested this on a openshift kubernetes cluster.

Hope it can help others.


Dockerfile.kubernetes
jaluma, line 1: Is there any limitation to have a separate `Dockerfile` for k8s?
It would be nice to merge your changes with the PR https://github.com/zylon-ai/private-gpt/pull/1792. What do you think?

Kraego: No there is no limitation it also works in local docker stack, it's just the additional layers to set the permission via group instead of a specific user, to run as rootless docker container. 
I've just created a separate dockerfile because there were already a `external` and `local` edition

jaluma: Can you try with last version? This includes GID and UID :)

Kraego: With GID and UID set the mounted volumes are writeable on kubernetes.

Kraego: It's not working on kubernetes. The UID to run the container is assigned by the node running the pod (https://www.redhat.com/en/blog/a-guide-to-openshift-and-uids). There is a kubernetes setting `securityContext/runAsUser`, but this feature is disabled on the cluster I'm working for safety concerns

jaluma: Is it related with openshift kubernetes?
I'm not sure if it's a native kubernetes feature or a Openshift one

Kraego: I‘ve looked it up. Its just a openshift thing (random UID). Vanilla kubernetes can run the docker container …

jaluma:
Why did you reopened this PR?

Kraego:
> Why did you reopened this PR?

I thought maybe as openshift variant. But if you like to close it, It's okay for me :-D

PR #1987, opened 2024-07-05
feat: bump dependencies

jaluma:
The current version has several outdated dependencies and does not allow the use of the latest changes from llama-index. This PR aims to update and prepare PrivateGPT to support the new changes.

### DIFF
--- a/private_gpt/components/vector_store/vector_store_component.py
+++ b/private_gpt/components/vector_store/vector_store_component.py
@@ -4,10 +4,10 @@ import typing
 from injector import inject, singleton
 from llama_index.core.indices.vector_store import VectorIndexRetriever, VectorStoreIndex
 from llama_index.core.vector_stores.types import (
+    BasePydanticVectorStore,
     FilterCondition,
     MetadataFilter,
     MetadataFilters,
-    VectorStore,
 )
 
 from private_gpt.open_ai.extensions.context_filter import ContextFilter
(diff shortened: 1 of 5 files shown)


private_gpt/components/vector_store/vector_store_component.py
```diff
@@ -4,10 +4,10 @@
 from injector import inject, singleton
 from llama_index.core.indices.vector_store import VectorIndexRetriever, VectorStoreIndex
 from llama_index.core.vector_stores.types import (
+    BasePydanticVectorStore,
```
imartinez, line 7: Why switching to BasePydanticVectorStore from VectorStore?

settings.yaml
```diff
@@ -78,7 +78,8 @@ nodestore:
   database: simple
 
 qdrant:
-  path: local_data/private_gpt/qdrant
```
imartinez, line 81: I'd keep the default setting using the disk-based local qdrant, instead of requiring a local qdrant service

settings.yaml
```diff
@@ -78,7 +78,8 @@ nodestore:
...
+  url: ${PGPT_QDRANT_URL:localhost:6333}
+  hybrid_search: ${PGPT_QDRANT_HYBRID_SEARCH:true}
```
imartinez, line 82: Let's do this in a different PR, given it is not directly related. Also, II don't think we are using PGPT_* env vars anywhere else. I'd say let's keep it simple and avoid env vars. Users can just override using their own settings_x.yaml

private_gpt/components/vector_store/vector_store_component.py
jaluma: Llama-index decision. Without that, mypy doesn't pass.
https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/vector_stores/types.py#L316

imartinez (approved):

PR #2008, opened 2024-07-23
feat: Add metadata to ingestion routes

NathanLenas:
# Description

This PR adds the ability to specify additional metadata in text and file ingestion routes.

This feature might be useful for people creating custom UIs with PrivateGPT and that want to add additional informations to the document sources, or for people using a custom ingestion pipeline with readers that may use the metadata. This feature could also be extended to bulk ingestion, but this would be a good start.

This is an important part of the project, so I wanted to make sure nothing breaks. 
I added two unit tests, one for the text ingestion route and one for the file ingestion route.
I also tested all of the ingestion modes to make sure nothing breaks. 

## Type of Change

Please delete options that are not relevant.
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests

**Test Configuration**:
* Firmware version: 0.5

## Checklist:
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [] Any dependent changes have been merged and published in downstream modules
- [x] I ran `make check; make test` to ensure mypy and tests pass

### DIFF
--- a/private_gpt/components/ingest/ingest_component.py
+++ b/private_gpt/components/ingest/ingest_component.py
@@ -40,11 +40,11 @@ class BaseIngestComponent(abc.ABC):
         self.transformations = transformations
 
     @abc.abstractmethod
-    def ingest(self, file_name: str, file_data: Path) -> list[Document]:
+    def ingest(self, file_name: str, file_data: Path, file_metadata : dict | None = None) -> list[Document]:
         pass
 
     @abc.abstractmethod
-    def bulk_ingest(self, files: list[tuple[str, Path]]) -> list[Document]:
+    def bulk_ingest(self, files: list[tuple[str, Path]], metadata : dict | None = None) -> list[Document]:
         pass
 
     @abc.abstractmethod
@@ -117,20 +117,20 @@ class SimpleIngestComponent(BaseIngestComponentWithIndex):
     ) -> None:
         super().__init__(storage_context, embed_model, transformations, *args, **kwargs)
 
-    def ingest(self, file_name: str, file_data: Path) -> list[Document]:
+    def ingest(self, file_name: str, file_data: Path, file_metadata : dict | None = None) -> list[Document]:
         logger.info("Ingesting file_name=%s", file_name)
-        documents = IngestionHelper.transform_file_into_documents(file_name, file_data)
+        documents = IngestionHelper.transform_file_into_documents(file_name, file_data, file_metadata)
         logger.info(
             "Transformed file=%s into count=%s documents", file_name, len(documents)
         )
         logger.debug("Saving the documents in the index and doc store")
         return self._save_docs(documents)
 
-    def bulk_ingest(self, files: list[tuple[str, Path]]) -> list[Document]:
+    def bulk_ingest(self, files: list[tuple[str, Path]], metadata : dict | None = None) -> list[Document]:
         saved_documents = []
         for file_name, file_data in files:
             documents = IngestionHelper.transform_file_into_documents(
-                file_name, file_data
+                file_name, file_data, metadata
             )
             saved_documents.extend(self._save_docs(documents))
         return saved_documents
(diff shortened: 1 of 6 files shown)


private_gpt/components/ingest/ingest_component.py
```diff
@@ -40,7 +40,12 @@ def __init__(
         self.transformations = transformations
 
     @abc.abstractmethod
-    def ingest(self, file_name: str, file_data: Path) -> list[Document]:
+    def ingest(
+        self,
+        file_name: str,
+        file_data: Path,
+        file_metadata: dict[str, str] | None = None,
```
jaluma, line 47: I would say it is better idea to type like `dict[str, Any] | None`

private_gpt/server/ingest/ingest_router.py
```diff
@@ -57,7 +73,11 @@ def ingest_file(request: Request, file: UploadFile) -> IngestResponse:
     service = request.state.injector.get(IngestService)
     if file.filename is None:
         raise HTTPException(400, "No file name provided")
-    ingested_documents = service.ingest_bin_data(file.filename, file.file)
+
+    metadata_dict = None if metadata is None else eval(metadata)
```
jaluma, line 77: I do not like how you're processing `metadata` field. Can you try to do a model or use a `dict`?
An example of function signature:
`def ingest_file(request: Request, file: UploadFile = File(...), metadata: Dict[str, str] = Body(...)) -> IngestResponse:`

tests/fixtures/ingest_helper.py
```diff
@@ -18,6 +19,18 @@ def ingest_file(self, path: Path) -> IngestResponse:
         ingest_result = IngestResponse.model_validate(response.json())
         return ingest_result
 
+    def ingest_file_with_metadata(self, path: Path, metadata: dict) -> IngestResponse:
+        files = {
+            "file": (path.name, path.open("rb")),
+            "metadata": (None, json.dumps(metadata)),
+        }
```
jaluma, line 26: Same as below. Ideally, it would say it has to be something like this:
```
files = {
  "file": (path.name, path.open("rb")),
   "metadata": metadata
}

**Review by jaluma — CHANGES_REQUESTED** [review]

File: private_gpt/components/ingest/ingest_component.py (hunk 1 above)
Comment (NathanLenas) [reply private_gpt/components/ingest/ingest_component.py:47]: This is something I thought about because of llama index's restriction for documents: the metadata can only be a flat value (string, int, float, see https://docs.llamaindex.ai/en/stable/module_guides/loading/documents_and_nodes/usage_documents/#metadata). By using string I made it "fool proof". But lmk if you want Any, i will change it

File: private_gpt/server/ingest/ingest_router.py (hunk 2 above)
Comment (NathanLenas) [reply private_gpt/server/ingest/ingest_router.py:77]: I tried to do it the way you suggested, but because the request contains a file upload, the request is handled differently and not processed as json, but as multipart/form-data, which prevents me from using a dict file type, I need to use a form field.

Comment (jaluma) [reply private_gpt/server/ingest/ingest_router.py:77]: What do you think about this proposal? It's a mock test, but it works. The only thing is that it takes `file` property as well.
```
async def extract_form_data(request: Request) -> Dict[str, Any]:
    form = await request.form()
    form_data = {key: value for key, value in form.items()}
    return form_data

@ingest_router.post("/ingest/file", tags=["Ingestion"])
def ingest_file(
    request: Request, file: UploadFile, metadata: dict[str, str] = Depends(extract_form_data)
) -> IngestResponse:
    ...
```

File: private_gpt/components/ingest/ingest_component.py (hunk 1 above)
Comment (jaluma) [reply private_gpt/components/ingest/ingest_component.py:47]: I believe it is better idea try to convert data when we're storing data into vector store that compromise all application with this decision. What do you think @imartinez?

File: private_gpt/server/ingest/ingest_router.py (hunk 2 above)
Comment (NathanLenas) [reply private_gpt/server/ingest/ingest_router.py:77]: This is a great suggestion, here is what i tried: 

```

async def extract_form_data(request: Request) -> Dict[str, Any] | None:
    form = await request.form()
    data = {key: value for key, value in form.items() if key != 'file'}
    return data["metadata"] if "metadata" in data else None

@ingest_router.post("/ingest/file", tags=["Ingestion"])
def ingest_file(
    request: Request, file: UploadFile, metadata: dict[str, Any] = Depends(extract_form_data)
) -> IngestResponse:
    
    service = request.state.injector.get(IngestService)
    
    if file.filename is None:
        raise HTTPException(400, "No file name provided") 
    
    print("metadata : " + str(metadata) + " type : " + str(type(metadata)))
    ingested_documents = service.ingest_bin_data(
        file.filename, file.file, metadata
    )
    return IngestResponse(object="list", model="private-gpt", data=ingested_documents)

```

The output from the print is metadata : {"foo":"bar"} type : <class 'str'>, so it is not intepreted as a dict. Digging deeper, it is not possible to have json encoded types such as a dict in a route that contains a File upload field ( see https://stackoverflow.com/questions/63110848/how-do-i-send-list-of-dictionary-as-body-parameter-together-with-files-in-fastap/71439821#71439821)

The only solutions I can think of is either to parse the string we receive as JSON, or to use eval().  
What's your opinion on these ?

jaluma: You can try to parse the string, it's better idea than `eval()` :)

PR #2031, opened 2024-08-01
fix: add progress bar when ollama is pulling models

jaluma:
# Description

Adding a progress bar when Ollama is downloading models

## Type of Change

- [X] New feature (non-breaking change which adds functionality)


imartinez (approved):
👏

PR #2044, opened 2024-08-05
fix: add cuda dockerfile + new docker-compose profile

jaluma:


jaluma:
Failure when starting the container
https://github.com/abetlen/llama-cpp-python/issues/1169#issuecomment-2228511452

MNLierman:
Is this working? Would love to use this in Docker on my 128 Cude Core Jetson.

PR #2047, opened 2024-08-06
chore: update matplotlib to 3.9.1-post1

jaluma:
Workaround: https://github.com/matplotlib/matplotlib/issues/28551


QuLogic:
Since 3.9.1 was yanked, it might not be necessary to have this pinned now; you shouldn't get 3.9.1 by default.

jaluma:
> Since 3.9.1 was yanked, it might not be necessary to have this pinned now; you shouldn't get 3.9.1 by default, though I'm not sure about poetry's locking. But there are 3.9.1.post1 wheels up now, which may be a better option.

I have updated this PR by upgrading only poetry.lock. We had upgraded to 3.9.1, and it's crashing when a windows user tries to install. Thank you for your reply anyway :)

imartinez (approved):

QuLogic:
> > Since 3.9.1 was yanked, it might not be necessary to have this pinned now; you shouldn't get 3.9.1 by default, though I'm not sure about poetry's locking. But there are 3.9.1.post1 wheels up now, which may be a better option.
> 
> I have updated this PR by upgrading only poetry.lock. We had upgraded to 3.9.1, and it's crashing when a windows user tries to install. Thank you for your reply anyway :)

I only meant the pinning in `pyproject.toml`; poetry still pulls the yanked version that it's locked to, so you'd still need to update the locks to grab v3.9.1.post1 instead.

jaluma:
> > > Since 3.9.1 was yanked, it might not be necessary to have this pinned now; you shouldn't get 3.9.1 by default, though I'm not sure about poetry's locking. But there are 3.9.1.post1 wheels up now, which may be a better option.
> > 
> > 
> > I have updated this PR by upgrading only poetry.lock. We had upgraded to 3.9.1, and it's crashing when a windows user tries to install. Thank you for your reply anyway :)
> 
> I only meant the pinning in `pyproject.toml`; poetry still pulls the yanked version that it's locked to, so you'd still need to update the locks to grab v3.9.1.post1 instead.

Yes, yes, that's why I removed the pinning reference in `myproject.toml`. Right now, I have downgraded the poetry.lock version, for frozen installations to 3.9.0. In any other situation, it will install the latest version 3.9.1-post1 instead

PR #2191, opened 2025-12-28
chore: add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

not sure about the fix in the test... CI would show?

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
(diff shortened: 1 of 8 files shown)


SMWQueens:
Stale pull request

yarikoptic:
yeap, nobody looked into it, afaik was generally ok, test logs are now gone but can redo if desired

AlfonsoLozana:

PR #2273, opened 2026-06-18
feat: control docs parser

AlfonsoLozana:
# Description

This PR adds a **vision-based extraction fallback** for document ingestion. When the primary parser fails to extract usable text from a PDF the ingest pipeline can now automatically retry extraction using a new `VisionReader` that rasterizes each page and uses a VLM to recover the content as Markdown.

**Motivation:** Some PDFs (scanned docs, non-standard font encodings) silently produce garbage text with the current extraction path. This change detects those failures and recovers the content instead of ingesting unusable data.

**How it works:**
1. Docling processes the PDF as usual.
2. If too many unmapped characters are detected (`glyph<UNKNOWN>` ratio above `failure_threshold`), it raises the new `ExtractionUnsuccessfulError`.
3. If `enable_vision_fallback=true`, `IngestComponent` catches the error and retries extraction with the `VisionReader` (rasterizes the PDF to images via `pypdfium2` and uses a VLM to extract text).
4. The parsing exception is only propagated if **both** methods fail.

The feature is **off by default** (`PGPT_ENABLE_VISION_FALLBACK=false`), so existing behavior is unchanged unless explicitly enabled.

### New settings
| Variable | Default | Controls |
|---|---|---|
| `enable_vision_fallback` (`PGPT_ENABLE_VISION_FALLBACK`) | `false` | Enables the VisionReader retry when the primary reader fails |
| `failure_threshold` (DoclingSettings) | `0.3` | Ratio of `glyph<UNKNOWN>` characters to consider extraction failed |
| `vision_documents` (TransformationSettings) | — | Sub-settings for the vision document pipeline (retries, semaphore, iterations) |
| `extraction_type_override` | `None` | Forces a concrete extraction strategy (`"mixed"`) without inference |

> Note: `VisionSettings.enable_evaluation` default changed from `true` to `false`.

### Dependencies
- `pypdfium2` (PDF rasterization)
- `pptx2md` (PPTX → image path)

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

Tested manually against real PDFs that previously failed extraction (CID-font PDFs producing `glyph<UNKNOWN>`), confirming that with `enable_vision_fallback=true` the VisionReader recovers the content and that with it disabled the original behavior is preserved. Also performed a manual self-review of the full diff.

- [ ] Added new unit/integration tests

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I ran `make check; make test` to ensure mypy and tests pass


private_gpt/components/multimodality/image_handler.py
```diff
@@ -169,11 +171,23 @@ async def infer_strategy(
         self, ctx: AnyContext, ev: WorkflowInitializedEvent
     ) -> StrategyInferredEvent | ImagesPreprocessedEvent:
 
-        iteration = await ctx.store.get("iteration", 1)
         enable_preprocessing = await ctx.store.get("enable_preprocessing")
+        extraction_type_override = await ctx.store.get("extraction_type_override")
         kwargs = await ctx.store.get("kwargs")
 
-        strategy = await self._infer_strategy(ev.image_blocks, seed=iteration, **kwargs)
+        if extraction_type_override:
```
jaluma: why?

private_gpt/settings/settings.py
```diff
@@ -1472,7 +1487,7 @@ class ReaderSettings(BaseModel):
 
 class VisionSettings(ReaderSettings):
     enable_evaluation: bool = Field(  # Preset
-        default=True,
```
jaluma, line 1475: why? you are changing not only this behaviour

jaluma (dismissed):

private_gpt/settings/settings.py
AlfonsoLozana: I've modified the default configuration so that it only affects the new pipeline

private_gpt/components/multimodality/image_handler.py
AlfonsoLozana: The idea was to avoid an inference, since processing long documents can be very costly with this extra inference. I've modified it to allow this step to take the value "none", so now you can simply skip this step

jaluma (approved):
LGTM!

jaluma (approved):

